PR #6, opened 2021-09-02
meta: add a whole bunch of skeleton files

woodruffw:
This sets up a basic development environment that should work well across both Linux and macOS.

The only external dependency is Python 3.6.


woodruffw:
Basic setup steps:

```bash
# if you don't already have Python 3.6
pyenv install 3.6.14
make dev
source env/bin/activate
```

tetsuo-cpp (approved):
Gave it a spin and it works great. LGTM!

PR #12, opened 2021-09-09
service: add `VulnerabilityService` interface type

### LINKED ISSUE #9: Design a generic adaptor for Python vulnerability services
Now that we've evaluated osv.dev (#2), we can start designing an abstract interface that individual vulnerability services can satisfy.

From discussion on Slack: we probably want something like this:

```python
class VulnerabilityService(ABC):
    def query(spec: Dependency) -> List[VulnerabilityResult]:
        ...

    def query_all(specs: List[Dependency]) -> Dict[Dependency, List[VulnerabilityResult]]:
        # naive implementation that can be overridden if
        # a particular service supports bulk queries
        for spec in specs:
            ...
```

tetsuo-cpp:
Closes #9

### DIFF
--- a/pip_audit/service/__init__.py
+++ b/pip_audit/service/__init__.py
@@ -1,5 +1,1 @@
-from pip_audit.service.service import (  # noqa: F401
-    Dependency,
-    VulnerabilityResult,
-    VulnerabilityService
-)
+from .interface import Dependency, VulnerabilityResult, VulnerabilityService  # noqa: F401
--- a/pip_audit/service/__init__.py
+++ b/pip_audit/service/__init__.py
@@ -1,1 +1,5 @@
-from service import Dependency, VulnerabilityResult, VulnerabilityService
+from pip_audit.service.service import (  # noqa: F401
+    Dependency,
+    VulnerabilityResult,
+    VulnerabilityService
+)
--- a/pip_audit/service/__init__.py
+++ b/pip_audit/service/__init__.py
@@ -1,0 +1,1 @@
+from service import Dependency, VulnerabilityResult, VulnerabilityService
--- a/pip_audit/service/service.py
+++ b/pip_audit/service/service.py
@@ -1,0 +1,31 @@
+from abc import ABC
+from dataclasses import dataclass
+from typing import Dict, List, Optional
+
+from packaging.version import Version
+
+
+@dataclass(frozen=True)
+class Dependency:
+    package: str
+    version: Version
+
+
+@dataclass(frozen=True)
+class VulnerabilityResult:
+    id: str
+    description: str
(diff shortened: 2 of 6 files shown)


tetsuo-cpp:
These are currently failing due to lack of code coverage. That makes sense, since this interface doesn't have an implementation yet.

pip_audit/service/service.py
```diff
@@ -0,0 +1,31 @@
+from abc import ABC
+from dataclasses import dataclass
+from typing import Dict, List, Optional
+
+from packaging.version import Version
+
+
+@dataclass(frozen=True)
```
tetsuo-cpp, line 8: I figured these would be handy since they generate some convenience methods and they're a form of documentation to indicate:
- That it's a POD type
- That it's immutable

It's not included in earlier Python versions though so I had to add it as a dependency. I can easily remove it if you'd like.

pip_audit/service/__init__.py
```diff
@@ -0,0 +1,5 @@
+from pip_audit.service.service import (  # noqa: F401
```
tetsuo-cpp, line 1: There's something weird going on with this file. If I do a `make lint`, it always ends up reformatting this file. I think something in the lint chain is modifying this file in a way that an earlier step doesn't agree with.

pip_audit/service/service.py
woodruffw: LGTM! I was also going to use dataclasses, for the same reasons 🙂 . Python 3.6 will be EOLed soon anyways, at which point we can remove the dep.

pip_audit/service/__init__.py
woodruffw: Yeah, this is probably my fault -- I need to tweak the `isort` configs to compensate for `black`'s preferences. I'll jump into your PR to do that.

woodruffw: Hmm, I wasn't able to repro -- `make lint` doesn't change anything for me unless I remove the `noqa`, which we should keep since we're doing a re-export here. Maybe a version mismatch?

pip_audit/service/service.py
```diff
@@ -0,0 +1,31 @@
...
+class Dependency:
+    package: str
```
woodruffw, line 10: Maybe overkill, but we might want to consider doing a `typing.NewType` for the `package` field in the future -- we expect package names to be consistent with PEP 423 (?), so having some sort of type that reflects that would probably be good.

woodruffw:
> These are currently failing due to lack of code coverage. That makes sense, since this interface doesn't have an implementation yet.

Yep! I'm gonna add some mocked tests to get us coverage for the interfaces.

PR #13, opened 2021-09-09
service: add OSV vulnerability service

### LINKED ISSUE #10: Support osv.dev via our service adaptor
As part of/following #9: we should support osv.dev via our adaptor.

tetsuo-cpp:
Closes #10


test/test_service.py
```diff
@@ -0,0 +1,25 @@
+from typing import List
+
+from packaging.version import Version
+
+import pip_audit.service as service
+
+
+def test_osv():
+    osv = service.OsvService()
```
tetsuo-cpp, line 9: What do you think about invoking the actual API as part of the test? Do you think it's ok.

I'll write a proper fixture and get those last few lines of coverage to get the build passing but I just wanted to run this by you first.

woodruffw: I think it's okay!

We can put the networked tests behind a `pytest.mark` + CLI option if we're concerned about the OSV API's reliability: https://stackoverflow.com/questions/47559524/pytest-how-to-skip-tests-unless-you-declare-an-option-flag

woodruffw:
(I pushed up the merge now that #12 is in; sorry for the churn).

pip_audit/service/osv.py
```diff
@@ -0,0 +1,50 @@
...
+        }
+        response: requests.Response = requests.post(
+            url=url,
+            data=json.dumps(query),
+        )
+
+        results: List[VulnerabilityResult] = []
+
+        # TODO(alex): Figure out what to do here
+        assert response.status_code == 200
```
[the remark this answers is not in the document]
tetsuo-cpp: Done! I've begun raising an `ServiceError` there and along with a unit test to check it.

test/test_service.py
tetsuo-cpp: Sounds good. I've intentionally avoided being too specific about what we get from the API to avoid failures if vulnerabilities get added to a package or modified, etc.

tetsuo-cpp:
Thanks for the fixes @woodruffw.
The test coverage should be up to par now.

woodruffw (approved):
LGTM!

PR #16, opened 2021-09-13
Add columns and JSON output formats

### LINKED ISSUE #5: Output formats for `pip-audit`
Apart from supporting a standard SBOM format (#3), `pip-audit` should have at least two output formats for the MVP:

* A human-readable format for display on a terminal or logs
* A JSON format for interpretation by other programs

tetsuo-cpp:
Closes #5

### DIFF
--- a/pip_audit/format/columns.py
+++ b/pip_audit/format/columns.py
@@ -1,0 +1,56 @@
+from itertools import zip_longest
+from typing import Any, Dict, Iterable, List, Tuple
+
+import pip_audit.service as service
+
+from .interface import VulnerabilityFormat
+
+
+def tabulate(rows: Iterable[Iterable[Any]]) -> Tuple[List[str], List[int]]:
+    """Return a list of formatted rows and a list of column sizes.
+    For example::
+    >>> tabulate([['foobar', 2000], [0xdeadbeef]])
+    (['foobar     2000', '3735928559'], [10, 4])
+    """
+    rows = [tuple(map(str, row)) for row in rows]
+    sizes = [max(map(len, col)) for col in zip_longest(*rows, fillvalue="")]
+    table = [" ".join(map(str.ljust, row, sizes)).rstrip() for row in rows]
+    return table, sizes
+
+
+class ColumnsFormat(VulnerabilityFormat):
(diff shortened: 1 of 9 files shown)


pip_audit/format/columns.py
```diff
@@ -0,0 +1,60 @@
+from itertools import zip_longest
+from typing import Any, Dict, Iterable, List, Tuple
+
+import pip_audit.service as service
+
+from .interface import VulnerabilityFormat
+
+
+def tabulate(rows: Iterable[Iterable[Any]]) -> Tuple[List[str], List[int]]:
```
tetsuo-cpp, line 9: I didn't think this was worth bringing in a dependency so I looked at how `pip list` does it and ripped the implementation out.

test/format/test_columns.py
```diff
@@ -0,0 +1,13 @@
+import pip_audit.format as format
+
+from .vuln_data import TEST_VULN_DATA
+
+
+def test_columns():
+    columns_format = format.ColumnsFormat()
+    expected_columns = """Package Version ID     Description              Affected Versions
+------- ------- ------ ------------------------ ------------------------------------
+foo     1.0     VULN-0 The first vulnerability  (0.9 => fix: 1.1), (N/A => fix: 1.4)
```
tetsuo-cpp, line 10: Wasn't sure how to best represent version ranges in a columnar format. Let me know if you have ideas!

pip_audit/format/columns.py
woodruffw: Works for me! cc @di for opinions.

If we do end up wanting to offload this to a dependency, https://github.com/astanin/python-tabulate is probably a good option.

test/format/test_columns.py
woodruffw: This seems reasonable to me (I think anything we do is going to be a bit shoehorned, since we're mapping a list of range into a single cell).

pip_audit/format/columns.py
di: This seems fine to me!

woodruffw (approved):

PR #22, opened 2021-09-13
pip_audit: dependency source skeleton

### LINKED ISSUE #8: Emit a warning when the underlying `pip` is sufficiently old
`pip-api` does a valiant job of wrapping `pip` for us, but certain functionality fundamentally degrades beyond a few versions:

* Earlier than `10.0.0b0`: `pip list -v --format=json` doesn't include location information, which hampers our ability to filter "system" dependencies (#7)

We should still try our hardest with these older `pip`s, but **also** emit a warning similar to the one emitted by `pip` itself when it's out of date.

### LINKED ISSUE #20: Develop a dependency collection interface
Similarly to how we've developed interfaces for vulnerability services and reporting formats: we should have an interface that individual sources of dependencies should implement.

That, in turn, will make our `Auditor` (#18) implementation sufficiently generic and extensible.

woodruffw:
Closes #20.
Closes #8.

### DIFF
--- a/pip_audit/dependency_source/pip.py
+++ b/pip_audit/dependency_source/pip.py
@@ -1,0 +1,49 @@
+"""
+Collect the local environment's active dependencies via `pip list`, wrapped
+by `pip-api`.
+"""
+
+import logging
+from typing import Iterator
+
+import pip_api  # type: ignore
+from packaging.version import Version
+
+from pip_audit.dependency_source import DependencySource, DependencySourceError
+from pip_audit.service import Dependency
+
+logger = logging.getLogger(__name__)
+
+# Versions of `pip` prior to this version don't support `pip list -v --format=json`,
+# which is our baseline for reliable output. We'll attempt to use versions before
+# this one, but not before complaining about it.
+_MINIMUM_RELIABLE_PIP_VERSION = Version("10.0.0b0")
+
+# NOTE(ww): The round-trip assignment here is due to type confusion: `pip_api.PIP_VERSION`
+# is a `Version` object, but it's a `pip_api._vendor.packaging.version.Version` instead
+# of a `packaging.version.Version`. Recreating the version with the correct type
+# ensures that our comparison operators work as expected.
+_PIP_VERSION = Version(str(pip_api.PIP_VERSION))
+
+
+class PipSource(DependencySource):
+    def __init__(self):
+        if _PIP_VERSION < _MINIMUM_RELIABLE_PIP_VERSION:
+            logger.warning(
+                f"Warning: pip {_PIP_VERSION} is very old, and may not provide reliable "
+                "dependency information! You are STRONGLY encouraged to upgrade to a "
+                "newer version of pip."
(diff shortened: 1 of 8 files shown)


woodruffw:
cc @di for feedback as well!

pip_audit/dependency_source/pip.py
```diff
@@ -0,0 +1,49 @@
...
+                "dependency information! You are STRONGLY encouraged to upgrade to a "
+                "newer version of pip."
+            )
+
+    def collect(self) -> Iterator[Dependency]:
+        # The `pip list` call that underlies `pip_api` could fail for myriad reasons.
+        # We collect them all into a single well-defined error.
+        try:
+            for (_, dist) in pip_api.installed_distributions().items():
+                yield Dependency(package=dist.name, version=Version(str(dist.version)))
```
tetsuo-cpp, line 43: I was wondering about:
```python
version=Version(str(dist.version))
```

It seems that the type of `dist.version` is `Union[LegacyVersion, Version]`. I'm wondering if maybe our `Dependency` type should use that instead of `Version`?

The parsing looks like this:
```python
def parse(version):
    # type: (str) -> Union[LegacyVersion, Version]
    """
    Parse the given version string and return either a :class:`Version` object
    or a :class:`LegacyVersion` object depending on if the given version is
    a valid PEP 440 version or a legacy version.
    """
    try:
        return Version(version)
    except InvalidVersion:
        return LegacyVersion(version)
```
So if it's a `LegacyVersion`, we'll end up raising in this spot too because we're trying to do the same thing.

di: I wouldn't worry about `LegacyVersion`, it is deprecated and will be removed soon: https://github.com/pypa/packaging/issues/321

tetsuo-cpp (approved):

pip_audit/dependency_source/pip.py
```diff
@@ -0,0 +1,49 @@
...
+# of a `packaging.version.Version`. Recreating the version with the correct type
+# ensures that our comparison operators work as expected.
+_PIP_VERSION = Version(str(pip_api.PIP_VERSION))
+
+
+class PipSource(DependencySource):
+    def __init__(self):
+        if _PIP_VERSION < _MINIMUM_RELIABLE_PIP_VERSION:
+            logger.warning(
+                f"Warning: pip {_PIP_VERSION} is very old, and may not provide reliable "
```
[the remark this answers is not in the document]
woodruffw: Excellent, thanks for catching and adding that.

pip_audit/dependency_source/pip.py
```diff
@@ -0,0 +1,49 @@
+"""
+Collect the local environment's active dependencies via `pip list`, wrapped
+by `pip-api`.
+"""
+
+import logging
+from typing import Iterator
+
+import pip_api  # type: ignore
```
woodruffw, line 9: #24 will address this.

PR #33, opened 2021-09-15
RC release prep

woodruffw:
Adds an appropriate classifier, and a release workflow.

### DIFF
--- a/.bump
+++ b/.bump
@@ -1,0 +1,3 @@
+[bump]
+input = pip_audit/version.py
+reset = true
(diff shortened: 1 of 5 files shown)


woodruffw:
I added a convenience `make` target that looks something like this:

```
make release
# or
make release BUMP_ARGS="--major"
```

...that target does everything up to the final `git push` steps, which it prints out for the user to trigger manually.

tetsuo-cpp (approved):

PR #40, opened 2021-09-17
Progress indicator

woodruffw:
Adds a `progress` based progress indicator to the CLI.

Some notes:

* I've subclasses `progress.spinner.Spinner` into `AuditSpinner` to make it behave a little better (the spin icon is on the LHS now, and we show each item being processed to give the user more feedback)
* I've confirmed that the default behavior of `progress` is reasonable when `stdout`/`stderr` aren't TTYs and that it doesn't interfere with our logging or other output

### DIFF
--- a/pip_audit/audit.py
+++ b/pip_audit/audit.py
@@ -47,8 +47,7 @@ class Auditor:
 
         if self._options.dry_run:
             # Drain the iterator in dry-run mode.
-            for spec in specs:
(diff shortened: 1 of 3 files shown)


woodruffw:
Demo:

[asciicast](https://asciinema.org/a/qDJAWEoiiWxEDPwjdZbIu8pbc)

tetsuo-cpp (approved):
LGTM!

PR #41, opened 2021-09-20
Bring in `resolvelib` based dependency resolver

### LINKED ISSUE #31: Evaluate `resolvelib`
As a parallel or perhaps complementary track to #29: [`resolvelib`](https://github.com/sarugaku/resolvelib) provides an abstract interface for resolving dependencies that are produced by a client-supplied "provider" (presumably something like us querying the PyPI APIs).

This might be too generic of a library for our use case, but it might be useful if we go further into reworking `pip-tools` to supply a Python API.

tetsuo-cpp:
Closes #31


woodruffw (approved):
LGTM! I'll push up a commit that adds appropriate attribution back to `resolvelib` for the provider we've based on their example.

PR #47, opened 2021-09-26
Add support for extras in `PyPIProvider`

### LINKED ISSUE #43: Add support for extras in `PyPIProvider`
At the moment, the provider that got from the `resolvelib` example, doesn't support extras in the package requirement. We should make this work.

Once we add support, we should ensure that [this unit test](https://github.com/trailofbits/pip-audit/blob/main/test/dependency_source/test_resolvelib.py#L47) works end-to-end.

tetsuo-cpp:
Closes #43

### DIFF
--- a/pip_audit/dependency_source/resolvelib/pypi_wheel_provider.py
+++ b/pip_audit/dependency_source/resolvelib/pypi_wheel_provider.py
@@ -119,16 +119,23 @@ class PyPIProvider(AbstractProvider):
 
     def find_matches(self, identifier, requirements, incompatibilities):
         requirements = list(requirements[identifier])
-        assert not any(r.extras for r in requirements), "extras not supported in this example"
 
         bad_versions = {c.version for c in incompatibilities[identifier]}
+
+        # TODO(alex): When do we have multiple requirements for a single dependency anyway?
+        #
+        # I'll need to figure out what situation this is modelling before I can decide whether it's
+        # ok to do this
+        extras = set()
+        for r in requirements:
+            extras |= r.extras
 
         # Need to pass the extras to the search, so they
         # are added to the candidate at creation - we
         # treat candidates as immutable once created.
         candidates = (
             candidate
-            for candidate in get_project_from_pypi(identifier, set())
+            for candidate in get_project_from_pypi(identifier, extras)
             if candidate.version not in bad_versions
             and all(candidate.version in r.specifier for r in requirements)
         )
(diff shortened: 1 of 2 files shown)


pip_audit/dependency_source/resolvelib/pypi_wheel_provider.py
```diff
@@ -119,16 +119,23 @@ def get_preference(self, identifier, resolutions, candidates, information):
 
     def find_matches(self, identifier, requirements, incompatibilities):
         requirements = list(requirements[identifier])
-        assert not any(r.extras for r in requirements), "extras not supported in this example"
 
         bad_versions = {c.version for c in incompatibilities[identifier]}
 
+        # TODO(alex): When do we have multiple requirements for a single dependency anyway?
```
tetsuo-cpp, line 125: Still reading through how `resolverlib` calls into this method to figure out the meaning of this.

tetsuo-cpp: I'm pretty sure accumulating extras like this is fine. The reason we can have multiple requirements for a single dependency is to accomodate other cases since `resolvelib` is a general purpose library.

The only issue I see is we have multiple requirements with different markers (say different package version for different Python versions). But I believe that since we added support for environment markers in `pip-api` which we use to read the requirements file, this should be a non-issue.

woodruffw: > The only issue I see is we have multiple requirements with different markers (say different package version for different Python versions). But I believe that since we added support for environment markers in pip-api which we use to read the requirements file, this should be a non-issue.

I _think_ this is right -- our marker support should smooth everything out, and everything that isn't smoothed over should be a genuine dependency conflict that we should handle gracefully.

> I tried adding some initial filtering for markers that evaluate to False but was having issues and didn't want to waste more time on it since pip-api should already be filtering this way. I've still got the changes stashed so if we think it's useful, I'm happy to add it back in and get it working.

Could you turn those changes into a patch and create a new issue with them + a short writeup of the problem? I agree that it's not worth investigating further given the work `pip-api` is already doing, but it'll be good to have a record of it just in case it proves to be an issue for unforeseen reasons.

tetsuo-cpp: > Could you turn those changes into a patch and create a new issue with them + a short writeup of the problem? I agree that it's not worth investigating further given the work pip-api is already doing, but it'll be good to have a record of it just in case it proves to be an issue for unforeseen reasons.

Done in #49

woodruffw (approved):
LGTM!

PR #48, opened 2021-09-27
Add support for source distributions in `PyPIProvider`

### LINKED ISSUE #44: Add support for source distributions in `PyPIProvider`
The current provider implementation only supports wheels, whereas we'll need to support source distributions. The relevant check is [here](https://github.com/trailofbits/pip-audit/blob/main/pip_audit/dependency_source/resolvelib/pypi_wheel_provider.py#L84).

@woodruffw had a useful comment [here](https://github.com/trailofbits/pip-audit/issues/31#issuecomment-920926216) outlining what this will entail.

tetsuo-cpp:
Closes #44

### DIFF
--- a/pip_audit/dependency_source/resolvelib/pypi_provider.py
+++ b/pip_audit/dependency_source/resolvelib/pypi_provider.py
@@ -11,6 +11,7 @@ from io import BytesIO
 from operator import attrgetter
 from platform import python_version
 from tarfile import TarFile
+from typing import List
 from urllib.parse import urlparse
 from zipfile import ZipFile
 
@@ -21,17 +22,19 @@ from packaging.specifiers import SpecifierSet
 from packaging.utils import canonicalize_name, parse_sdist_filename, parse_wheel_filename
 from packaging.version import Version
 from resolvelib.providers import AbstractProvider
+from virtualenvapi.manage import VirtualEnvironment  # type: ignore
 
 PYTHON_VERSION = Version(python_version())
 
 
 class Candidate:
-    def __init__(self, name, version, url=None, extras=None, is_wheel=True):
+    def __init__(self, name, version, url=None, extras=None, is_wheel=True, filename=str()):
         self.name = canonicalize_name(name)
         self.version = version
         self.url = url
         self.extras = extras
         self.is_wheel = is_wheel
+        self.filename = filename
 
         self._metadata = None
         self._dependencies = None
@@ -103,7 +106,7 @@ def get_project_from_pypi(project, extras):
 
         # TODO: Handle compatibility tags?
 
-        yield Candidate(name, version, url=url, extras=extras, is_wheel=is_wheel)
+        yield Candidate(name, version, url=url, extras=extras, is_wheel=is_wheel, filename=filename)
 
 
 def get_metadata_for_wheel(url):
@@ -120,13 +123,27 @@ def get_metadata_for_wheel(url):
 
 def get_metadata_for_sdist(url):
     data = requests.get(url).content
-    with TarFile.open(fileobj=BytesIO(data), mode="r:gz") as t:
-        for n in t.getnames():
-            if n.endswith("PKG-INFO"):
-                p = BytesParser()
-                return p.parse(t.extractfile(n), headersonly=True)
 
-    return EmailMessage()  # pragma: no cover
+    # Extract archive onto the disk
+    with TarFile.open(fileobj=BytesIO(data), mode="r:gz") as t:
+        t.extractall()
+
+    import os
+
+    # TODO(alex): Substitute the directory name here
+    cwd = os.getcwd()
+    pkg_path = os.path.join(cwd, "ansible-core-2.11.5")
+    ve = VirtualEnvironment("test_env/")
+    ve.install(f"-e {pkg_path}")
+
+    reqs: List[Requirement] = []
+    for name, version in ve.installed_packages:
+        if name.startswith("-e"):
+            continue
+        reqs.append(Requirement(f"{name}=={version}"))
+
+    # Just to keep things working
+    return EmailMessage()
 
 
 class PyPIProvider(AbstractProvider):
--- a/pip_audit/dependency_source/resolvelib/pypi_provider.py
+++ b/pip_audit/dependency_source/resolvelib/pypi_provider.py
@@ -1,5 +1,5 @@
 """
-A (wheel-only) `resolvelib` provider implementation that resolves against PyPI.
+A `resolvelib` provider implementation that resolves against PyPI.
 
 Closely adapted from `resolvelib`'s examples, which are copyrighted by the `resolvelib`
 authors under the ISC license.
--- a/pip_audit/dependency_source/resolvelib/pypi_provider.py
+++ b/pip_audit/dependency_source/resolvelib/pypi_provider.py
@@ -174,7 +174,8 @@ class PyPIProvider(AbstractProvider):
             if candidate.version not in bad_versions
             and all(candidate.version in r.specifier for r in requirements)
         )
-        return sorted(candidates, key=attrgetter("version"), reverse=True)
+        # We want to prefer more recent versions and prioritize wheels
+        return sorted(candidates, key=attrgetter("version", "is_wheel"), reverse=True)
 
     def is_satisfied_by(self, requirement, candidate):
         if canonicalize_name(requirement.name) != candidate.name:
--- a/pip_audit/dependency_source/resolvelib/pypi_provider.py
+++ b/pip_audit/dependency_source/resolvelib/pypi_provider.py
@@ -6,13 +6,13 @@ authors under the ISC license.
 """
 
 import os
-import shutil
 from email.message import EmailMessage
 from email.parser import BytesParser
 from io import BytesIO
 from operator import attrgetter
 from platform import python_version
 from tarfile import TarFile
+from tempfile import TemporaryDirectory
 from typing import Set
 from urllib.parse import urlparse
 from zipfile import ZipFile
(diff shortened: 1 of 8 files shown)


tetsuo-cpp (commented):
I'm still trying to find an sdist only package that has dependencies in the `PKG-INFO`. As it stands, the `ansible-core` test is only testing the filename parsing since it doesn't have package dependencies.

As soon as I find one, I'll add another unit test to check that the dependencies get read out the metadata properly.

woodruffw:
> I'm still trying to find an sdist only package that has dependencies in the PKG-INFO. As it stands, the ansible-core test is only testing the filename parsing since it doesn't have package dependencies.

Yeah, this is concerning -- I was under the impression that building a source distribution would include the dependencies in the `PKG-INFO`, but it doesn't look like it does for any of the packages I've checked either. I was aware `setup.py` was executed on the client side, so I guess that means dependency resolution happens wholesale at that point as well.

cc @di for thoughts: `pip-audit` probably shouldn't run `setup.py` from arbitrary dependencies, since dependencies can put all kinds of system-clobbering nonsense in their setup scripts (much less the problem of malicious ones). Perhaps we could support source distributions opportunistically (if whatever tool constructed them includes the correct metadata in the `PKG-INFO`), and otherwise warn the user that a dependency isn't being fully resolved?

pip_audit/dependency_source/resolvelib/pypi_provider.py
```diff
@@ -81,21 +87,46 @@ def get_project_from_pypi(project, extras):
...
+            name, version = parse_filename(filename, is_wheel)
         except InvalidVersion:
             # Ignore files with invalid versions
             continue
 
-        yield Candidate(name, version, url=url, extras=extras)
+        yield Candidate(name, version, url=url, extras=extras, is_wheel=is_wheel)
+
+
+def parse_filename(filename: str, is_wheel: bool) -> Tuple[str, Version]:
```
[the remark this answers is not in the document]
woodruffw: `packaging` provides `parse_wheel_filename` and `parse_sdist_filename`, so we'll swap those in! All of the PyPA tooling uses those (IIRC), so they're well-tested.

di:
> I was under the impression that building a source distribution would include the dependencies in the PKG-INFO, but it doesn't look like it does for any of the packages I've checked either.

Yes, this is true for 99% of source distributions on PyPI, see https://dustingram.com/articles/2018/03/05/why-pypi-doesnt-know-dependencies/

It _is_ possible to do with build backends that support [PEP 621](https://www.python.org/dev/peps/pep-0621/) (see [spam-1.2.3.tar.gz](https://github.com/trailofbits/pip-audit/files/7238218/spam-1.2.3.tar.gz) for example) but is still not very common in practice.

We'll probably need to do the same thing that other tools do to work around this, either a) install it in a vacuum and see what else gets installed, or b) attempt to build it into a wheel and see what that wheel declares as subdependencies.

woodruffw:
> We'll probably need to do the same thing that other tools do to work around this, either a) install it in a vacuum and see what else gets installed, or b) attempt to build it into a wheel and see what that wheel declares as subdependencies.

Gotcha. Doing package installation in a truly isolated environment is going to be challenging across platforms. Off the top of my head:

* Virtual environments: not a security boundary; too easy to escape with a malicious or badly written `setup.py`
* Docker/other containers: host mismatch concerns, **extremely** weak security boundaries, not a reasonable runtime assumption for all users
* Virtualization/VMs: same host mismatch concerns, reasonable security boundaries, but also not a reasonable runtime assumption

That being said, perhaps all of those are too paranoid: users are using `pip-audit` presumably because they intend to eventually install the audited dependencies, and so anything bad that _would_ happen in a porously isolated environment will presumably happen later anyways/as well.

Perhaps we flag this behavior, where `-r` treats every resolved `sdist` as a leaf by default and only follows them if something like `--insecure-follow-sdists` is also passed? Thoughts?

di:
I think we're confusing the threats here a bit. There's two distinct ones:

1. A legitimate package has a security vulnerability we want to know about;
2. A malicious package might be installed and do malicious things during installation.

This tool is focused on the first threat, not the second. This assumes that a given package itself is not malicious, and thus can be installed, but using it in a certain way might introduce a vulnerability.

Anything involving the second threat is not currently this project's responsibility: it can assume that by the nature of the project existing on PyPI, it is not malicious. (Obviously this is not always true, but without some ability to detect malicious packages in real time, that's as good as we can get).

woodruffw:
Yep; I definitely don't want to make item (2) this project's responsibility!

I'm mostly trying to exercise diligence over our inputs/make sure we have an explanation for a foreseeable and reasonable question, i.e. "why is `pip-audit` running arbitrary code while auditing my dependencies?" Your blog post answers that neatly and explains exactly why the Python packaging ecosystem is structured the way it is, but people coming from other packaging ecosystems with static dependency resolution might find it surprising.

That being said, we'll go ahead with an isolation approach that assumes the trustworthiness of packages resolved through PyPI.

pip_audit/dependency_source/resolvelib/pypi_provider.py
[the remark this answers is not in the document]
tetsuo-cpp: Thanks @woodruffw!

pip_audit/dependency_source/resolvelib/pypi_provider.py
```diff
@@ -110,6 +121,34 @@ def get_metadata_for_wheel(url):
...
+    data = requests.get(url).content
+
+    # Extract archive onto the disk
+    with TarFile.open(fileobj=BytesIO(data), mode="r:gz") as t:
+        # The directory is the first member in a tarball
+        names = t.getnames()
+        pkg_dir = names[0]
+        t.extractall()
+
+    cwd = os.getcwd()
```
tetsuo-cpp, line 134: I'm planning to add some code to generate an unused directory name and extract the tarball and put the virtualenv in there just so we don't get clashes if the user just so happens to have something called `tmp_env` in their cwd.

But this is the general gist of what I'm intending to do. I'm making a fresh virtualenv, installing the sdist on it and then doing a `pip freeze` to figure out what packages got installed. One issue (which I think is pretty minor) is that this code is reporting back some seed packages like `setuptools`, `pip`, etc.

woodruffw: This general approach looks reasonable to me!

Instead of building the virtualenv in the current directory, we should probably use the system temporary directory by way of the `tempfile` module. In particular, we can do some naive key-style caching by using a scheme like this:

```
${TMPDIR}/pip-audit-${some-hash-of-requirements-input}/
```

...and if the hash hasn't changed, we can assume that the state under that directory is up-to-date and not recompute it. That in turn involves leaving temporary directories around, however, so maybe it isn't worth it.

pip_audit/dependency_source/resolvelib/pypi_provider.py
```diff
@@ -18,19 +21,21 @@
 import requests
 from packaging.requirements import Requirement
 from packaging.specifiers import SpecifierSet
-from packaging.utils import canonicalize_name
-from packaging.version import InvalidVersion, Version
+from packaging.utils import canonicalize_name, parse_sdist_filename, parse_wheel_filename
+from packaging.version import Version
 from resolvelib.providers import AbstractProvider
+from virtualenvapi.manage import VirtualEnvironment  # type: ignore
```
woodruffw, line 27: We probably want to use the builtin `venv` module provided by Python, since "venv" is the standard solution as of Python 3 and doesn't make as many legacy (i.e. Python 2) accommodations. I'll work that out.

tetsuo-cpp: I did see that module but couldn't spot a way to inspect packages and things like that. It seemed to be more about creating virtual envs from what I could tell.

woodruffw: Yeah, it looks like the public API is pretty barebones. Hmm. We could probably do something like entering the virtual environment and then using `pip-api`, but that's a little circular, so maybe sticking with `virtualenvapi` makes sense for now 🙂

pip_audit/dependency_source/resolvelib/pypi_provider.py
tetsuo-cpp: Using temporary directories in 66e6c7cb170483273c1bf96c828b49009cb3c31a.

I decided to just keep it simple for now and have it use temporary directories for the sdist extraction and the virtual environment. I figure that if performance turns out to be a problem for real life `requirements.txt`, then we can introduce caching at a later point.

pip_audit/dependency_source/resolvelib/pypi_provider.py
```diff
@@ -111,6 +122,37 @@ def get_metadata_for_wheel(url):
...
+        pkg_path = os.path.join(pkg_dir, pkg_name)
+
+        with TemporaryDirectory() as ve_dir:
+            ve = VirtualEnvironment(ve_dir)
+            ve.install(f"-e {pkg_path}")
+
+            for name, version in ve.installed_packages:
+                # Skip the editable package since that's the one we're finding dependencies for
+                #
+                # The `virtualenvapi` package also has a bug where comments in the `pip freeze`
```
tetsuo-cpp, line 147: We can remove this check once https://github.com/sjkingo/virtualenv-api/pull/54 gets merged in.

pip_audit/dependency_source/resolvelib/pypi_provider.py
tetsuo-cpp: Hmm, yeah maybe that is a better idea. I'm starting to wonder if it's a good idea to add `virtualenvapi` as a dependency. 

The part that I'm not sure about is "entering the virtual environment". Seems like I'd have to call into a separate Python via `system` or something like that. Is that your understanding too?

woodruffw: > The part that I'm not sure about is "entering the virtual environment". Seems like I'd have to call into a separate Python via system or something like that. Is that your understanding too?

There's a mysterious `context` object in the `venv` API, and I was hoping that something in there would give us access to the temporary environment so that we don't have to do anything roundabout like shelling out to a child Python process. But I haven't had time to look into it yet; could you spend some time investigating that and seeing whether there's some nicety there that we can lean on?

Otherwise: yeah, I think building up a bit of scaffolding for creating a Python subprocess within that virtual environment is the right approach.

pip_audit/dependency_source/resolvelib/pypi_provider.py
```diff
@@ -111,6 +122,37 @@ def get_metadata_for_wheel(url):
     return EmailMessage()  # pragma: no cover
 
 
+def get_metadata_for_sdist(url):
+    data = requests.get(url).content
```
woodruffw, line 126: This HTTP request can fail, so let's do `raise_for_status` or something similar to propagate it as an exception that we catch and transform into one of our own exceptions.

tetsuo-cpp: da69505d7d5c657c6e28ae69f605693e8067f06d

pip_audit/dependency_source/resolvelib/pypi_provider.py
tetsuo-cpp: I don't see anything like that in the `venv` API. There is some code [here](https://github.com/python/cpython/blob/3.9/Lib/venv/__init__.py#L111) that shows how the context gets created, but it doesn't seem to have what we want. From what I can see, the `venv` only really wraps what you can do with `venv` from the command line. So anything like introspecting environments, installing things within them, etc looks to be out of scope for `venv`.

tetsuo-cpp: I think if we want to remove the `virtualenvapi` dependency, we'll have to write some code that does:
- Creating the virtual environment with `venv`.
- Shell out to Python and do something like:
```
source env/bin/activate
pip install -e sdist_dir/
pip install pip-api
python -c "import pip_api; print(pip_api.installed_packages)"
```
- Parse the shell output into package name and version pairs.
- Filter out `pip_api` dependencies since those will turn up in the output also.

This is all possible but I'm thinking that it might be worth just getting everything working end to end and then having "remove `virtualenvapi` dependency" as a nice-to-have just so we can unblock the other tasks and keep moving.

What do you think @woodruffw.

woodruffw: Yeah, let's put this info in an issue and revisit it once we have this initial version in.

tetsuo-cpp: Captured this in #50.

woodruffw: Thanks!

woodruffw (approved):

PR #51, opened 2021-10-04
Add a `RequirementSource` to our dependency source API

### LINKED ISSUE #23: Add a `RequirementSource` to our dependency source API.
#22 will implement `DependencySource` for `pip list`; we need a corresponding implementation for parsing and concretizing requirement files.

tetsuo-cpp:
Closes #23

### DIFF
--- a/pip_audit/dependency_source/requirement.py
+++ b/pip_audit/dependency_source/requirement.py
@@ -5,7 +5,12 @@ from packaging.requirements import Requirement
 from pip_api import parse_requirements
 from pip_api._parse_requirements import UnparsedRequirement
 
-from pip_audit.dependency_source import DependencyResolver, DependencySource
+from pip_audit.dependency_source import (
+    DependencyResolver,
+    DependencyResolverError,
+    DependencySource,
+    DependencySourceError,
+)
 from pip_audit.service import Dependency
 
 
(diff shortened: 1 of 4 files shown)


tetsuo-cpp:
I'll fill the remaining coverage tomorrow, but this is the general idea.

pip_audit/dependency_source/requirement.py
```diff
@@ -0,0 +1,42 @@
...
+from pip_audit.service import Dependency
+
+
+class RequirementSource(DependencySource):
+    def __init__(self, filenames: List[Path], resolver: DependencyResolver):
+        self.filenames = filenames
+        self.resolver = resolver
+
+    def collect(self) -> Iterator[Dependency]:
+        # TODO(alex): I wonder whether we need to do some deduplication of requirements/dependencies
```
tetsuo-cpp, line 23: I have an idea that if two requirements files share the same dependency, it might appear twice in the output and get audited twice.

If the versions are different, I think you'd want two separate entries since they can suffer from different vulnerabilities. But in the case of identical versions, it might be worth deduplicating them.

woodruffw: Yeah, deduplicating them makes sense.

I wonder if it also make sense to warn or otherwise signal users when they have two different (and incompatible) versions for the same dependency -- actually installing those into the same (virtual) environment would be a conflict, one that `pip-audit` manages to avoid circumstantially (by using a different venv for each requirement source). 

But that's perhaps too far out of scope, so we can ignore it for now 🙂

PR #54, opened 2021-10-08
Remove `virtualenv-api` dependency

### LINKED ISSUE #50: Remove `virtualenv-api` dependency
In #48, I introduced a dependency to `virtualenv-api` to install source distributions in a virtual environment and figure out what their dependencies are.

We should remove this dependency and instead rely on builtin modules. The builtin `venv` module has less functionality (it's more about creating virtual environments than modifying or introspecting them) so we'll have to do some legwork to replace it. Specifically:

- Creating the virtual environment with `venv`.
- Shell out to Python and do something like:
```
source env/bin/activate
pip install -e sdist_dir/
pip install pip-api
python -c "import pip_api; print(pip_api.installed_packages)"
```
- Parse the shell output into package name and version pairs.
- Filter out `pip_api` dependencies since those will turn up in the output also.

tetsuo-cpp:
Closes #50

### DIFF
--- a/pip_audit/cli.py
+++ b/pip_audit/cli.py
@@ -47,11 +47,11 @@ class OutputFormatChoice(str, enum.Enum):
     Columns = "columns"
     Json = "json"
 
-    def to_format(self) -> VulnerabilityFormat:
+    def to_format(self, output_desc: bool) -> VulnerabilityFormat:
         if self is OutputFormatChoice.Columns:
-            return ColumnsFormat()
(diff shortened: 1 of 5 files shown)


tetsuo-cpp:
@woodruffw
Thanks!

I can't seem to repro that (maybe because my system Python version is different). Could you give it a try now and check if my latest commit help?

Also, I've made #60 to capture the issue about descriptions in the output. I'll follow up with a separate PR to add a command line flag to toggle that.

Issue #56, opened 2021-10-08
Make the CLI more responsive when auditing from requirements files


woodruffw:
Currently, the CLI can block without any indicators for a decent period while collecting dependency information from requirements files.

We should figure out a way to make it responsive (or provide a working indicator) during that initial period.

woodruffw:
Instead of trying to parallelize our dependency collection (and therefore faster), let's figure out what additional state we can introduce to make the CLI more responsive when resolving depenencies.

Some thoughts:

* I'm not as familiar with this pattern in Python, but a pattern I've used in other languages is to pass a callback or other hook object into the "actor" object. The "actor" can then call the hook for each "action" taken, which in our case would be resolving each sub-dependency. This allows our CLI indicator of choice (spinner, whatever) to synchronize with the internals of the dependency resolver. 
    * Pros: Easy to implement, easy to write in a fashion that doesn't violate the boundary between `pip-audit`'s APIs and CLI.
    * Cons: Assuming that we're not using `async` in Python (or running the CLI in another thread), this effectively makes the CLI's update work part of the main work for dependency resolution. Runtime is probably still dominated by `resolvelib` + our venv work, but it's a weak anti-pattern.

PR #64, opened 2021-10-13
Add spinner for dependency collection

tetsuo-cpp:
This doesn't fully address #56 but it's low hanging fruit that fits cleanly in our current design. The response from the CLI is a bit "bursty" since we're updating when returning dependencies instead of prior to dependency resolution. But for requirements files with a decent number of requirements, this is still a lot better than what we currently have.

I think ideally, the CLI would also signal that its resolving dependencies for a requirement. Also, parallelising sdist resolution will influence what information we can return back to the CLI. So I figure we can use this patch as a stopgap to give us more reasonable behaviour while we figure out how this should work in the long term.

### DIFF
--- a/pip_audit/service/osv.py
+++ b/pip_audit/service/osv.py
@@ -23,8 +23,10 @@ class OsvService(VulnerabilityService):
         results: List[VulnerabilityResult] = []
 
         # Check for an unsuccessful status code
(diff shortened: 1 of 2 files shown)


woodruffw:
Sounds good!

Issue #53, opened 2021-10-06
Provide example schema for PyPI vulnerability service


tetsuo-cpp:
The current plan is to extend the [release endpoint](https://warehouse.pypa.io/api-reference/json.html#release) to provide vulnerability information for a given release. We should provide an example schema that has all the information that we need for `pip-audit`.

di:
This is probably going to look more like:

```json
  "vulns": {
    "id": "PYSEC-1",
    "details": "The first vulnerability",
    "link": "https://osv.dev/vulnerability/PYSEC-1",
    "aliases": [
      "CVE-2020-36242"
    ],
    "ranges": [
      {
        "introduced": "3.1",
        "fixed": "3.3.2"
      }
    ]
  }
```
Will that work?

di:
Blocked on https://github.com/google/osv/pull/238

tetsuo-cpp:
Hmm, actually. If a vulnerability affects multiple versions, is there going to be a duplicate vulnerability entry for every affected version?

I wonder whether the responses are going to become very large for certain projects. Not only is it going to fetch vulnerability information for every version that's ever been published, but the same vulnerability information is going to be duplicated using this schema. The way I imagined it working was that you'd get vulnerabilities affecting only the version that you're querying with `/pypi/<project_name>/<version>/json`.

tetsuo-cpp:
Thanks @di. That format definitely works for our purposes.

tetsuo-cpp:
Awesome, thanks @di!

tetsuo-cpp:
Closing this issue since this is more about deciding on a schema. The work to adding support for the new PyPI service is tracked in #65.

PR #65, opened 2021-10-14
pip_audit: Add PyPI vulnerability service

### LINKED ISSUE #11: Support PyPI's vulnerability service via our service adaptor
Following #9 and #10: once it's available, we should support PyPI's vulnerability service via our adaptor.

tetsuo-cpp:
Closes #11

### DIFF
--- a/test/service/test_pypi.py
+++ b/test/service/test_pypi.py
@@ -1,5 +1,7 @@
 from typing import List
 
+import pytest
+import requests
 from packaging.version import Version
 
 import pip_audit.service as service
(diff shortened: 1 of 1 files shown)


tetsuo-cpp:
This is proof of concept code based on our discussion on #53 about the new PyPI API. When the new API gets rolled out, I'll add some tests and we can get this in.

tetsuo-cpp:
Hmm, this can't really be tested properly until the API changes are rolled out but I've been thinking about how this is going to work.

Since the PyPI API is using OSV as a backend, it might make sense to just parameterise the OSV tests and test that the implementations return the same results. The only bits that might be nasty is figuring out what methods to monkey patch in each case since the implementations are different. Also, the OSV backend isn't able to tell whether a package is unknown since it won't know about packages/releases that have no recorded vulnerabilities whereas, the PyPI backend will raise on any package/release combination that it doesn't know about. So there are some slight behavioural differences between the two.

tetsuo-cpp:
@woodruffw Ok, the PyPI vulnerability service is online now. So this is good to go.

PR #95, opened 2021-11-01
pip_audit: Update state feedback when fetching metadata

### LINKED ISSUE #89: Make the CLI *even more* responsive
Follow-up on #56, per https://github.com/trailofbits/pip-audit/pull/87#issuecomment-954320210:

> Regarding the feedback not being great at the beginning, in the future it might be worth reporting some extra state in the PyPIProvider. Something like:

> `Gathering metadata for sdist {name} ({version})`

tetsuo-cpp:
Closes #89

- [x] Update state for virtual env creation.
- [x] Update state for sdist installation

### DIFF
--- a/pip_audit/dependency_source/resolvelib/pypi_provider.py
+++ b/pip_audit/dependency_source/resolvelib/pypi_provider.py
@@ -133,7 +133,7 @@ class Candidate:
             pkg_path = os.path.join(pkg_dir, pkg_name)
 
             with TemporaryDirectory() as ve_dir:
-                ve = VirtualEnv(["-e", pkg_path])
+                ve = VirtualEnv(["-e", pkg_path], self.state)
                 ve.create(ve_dir)
 
                 if self.state is not None:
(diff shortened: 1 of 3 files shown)


pip_audit/dependency_source/resolvelib/pypi_provider.py
```diff
@@ -75,8 +87,68 @@ def dependencies(self):
             self._dependencies = list(self._get_dependencies())
         return self._dependencies
 
+    def _get_metadata_for_wheel(self):
```
tetsuo-cpp, line 90: Not really any code changes here apart from the `update_state` lines. I just moved them to methods of the `Candidate` since I'm using the package information in the feedback messages.

tetsuo-cpp:
It does still spend a chunk of time in the sdist installation, so it's not perfect. But I'd say that this is about as good as it's going to get.

woodruffw:
This might be a bug in my local environment, but the progress spinner is now broken for me:

[asciicast](https://asciinema.org/a/446637)

Will debug.

woodruffw:
Hmm, the regression starts with 6f36c45157ecf0a0fc8a0e4cd30de48b4d83c408, which doesn't have anything particularly suspicious in it.

woodruffw:
Ignore all of that: iTerm renders the spinner correctly, so it's something weird with Alacritty or my termcap.

woodruffw (approved):
LGTM!

PR #104, opened 2021-11-03
cli: Add option to toggle progress spinner

### LINKED ISSUE #96: Allow users to explicitly disable the progress spinner
The spinner is enabled when `stderr` is a TTY by default. We should allow users to explicitly disable the spinner even when their standard error stream is going to a TTY, since not all TTYs behave well with line-pinned spinners and not all users may want a cute progress indicator at all.

tetsuo-cpp:
Closes #96

### DIFF
--- a/pip_audit/cli.py
+++ b/pip_audit/cli.py
@@ -7,6 +7,7 @@ import enum
 import logging
 import os
 import sys
+from contextlib import ExitStack
 from pathlib import Path
 from typing import List, Optional
 
@@ -88,6 +89,22 @@ class VulnerabilityDescriptionChoice(str, enum.Enum):
         return self.value
 
 
+@enum.unique
+class ProgressSpinnerChoice(str, enum.Enum):
+    """
+    Whether or not `pip-audit` should display a progress spinner.
+    """
+
+    On = "on"
+    Off = "off"
+
+    def to_bool(self) -> bool:
+        return self is ProgressSpinnerChoice.On
+
+    def __str__(self):
+        return self.value
+
+
 def audit():
     """
     The primary entrypoint for `pip-audit`.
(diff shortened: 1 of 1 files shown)


pip_audit/cli.py
```diff
@@ -147,6 +164,12 @@ def audit():
         type=Path,
         help="the directory to use as an HTTP cache for PyPI; uses the `pip` HTTP cache by default",
     )
+    parser.add_argument(
```
tetsuo-cpp, line 167: I decided to stick with a similar interface to `pip` instead of having a `--no-progress` flag.

pip_audit/cli.py
```diff
@@ -155,7 +178,9 @@ def audit():
     output_desc = args.desc.to_bool(args.format)
     formatter = args.format.to_format(output_desc)
 
-    with AuditSpinner() as state:
+    with ExitStack() as stack:
+        state = stack.enter_context(AuditSpinner()) if args.progress_spinner.to_bool() else None
```
tetsuo-cpp, line 182: This is necessary in order to use the scoped enter/exit approach conditionally. I think it's still useful to make sure we `finalize`, but I could be convinced otherwise.

pip_audit/cli.py
```diff
@@ -88,6 +89,22 @@ def __str__(self):
...
+class ProgressSpinnerChoice(str, enum.Enum):
+    """
+    Whether or not `pip-audit` should display a progress spinner.
+    """
+
+    On = "on"
+    Off = "off"
+
+    def to_bool(self) -> bool:
+        return self is ProgressSpinnerChoice.On
```
woodruffw, line 102: Let's use `__bool__` for this, since `bool(...)` uses it under the hood.

pip_audit/cli.py
woodruffw: TIL about `ExitStack`! Perhaps we could push this one level further -- instead of having `AuditState` be an ABC, it could manage an `ExitStack` whose members conform to an ABC/Subtype protocol. I'll fiddle around with that, but this LGTM for now.

woodruffw (approved):

PR #106, opened 2021-11-04
CI, Makefile, README: enforce `pip-audit --help` changes

### LINKED ISSUE #105: CI: Enforce up-to-datedness of `pip-audit --help` in the README
This should be very easy to do with a GitHub action that just `diff`s that section of the README against `pip-audit --help`.

woodruffw:
Closes #105.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -35,6 +35,11 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       - uses: actions/setup-python@v2
+        # NOTE(ww): Important: use pip-audit's minimum supported Python version
+        # in this check, since Python can change the `--help` rendering in
+        # `argparse` between major versions.
+        with:
+          python-version: "3.6"
       - name: deps
         run: make dev
       - name: check-readme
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -30,3 +30,25 @@ jobs:
...
+      - name: check-readme
+        run: |
+          diff \
+            <( \
+              awk '/@begin-pip-audit-help@/{f=1;next} /@end-pip-audit-help@/{f=0} f' \
+                < README.md | sed '1d;$d' \
+            ) \
+            <( \
+              make run ARGS="--help" \
+            )
```
woodruffw, line 54: This is a bit of a doozy, so the summary:

1. We run `diff` with two here-files:
1. The first is the contents of `README.md` filtered down to just the block that has the `pip-audit --help` text in it
2. The second is the result of calling `pip-audit --help` (with the latest changes)

tetsuo-cpp: Nifty! I really need to get around to learning some `awk` one of these days.

tetsuo-cpp (approved):
LGTM!

PR #129, opened 2021-11-10
github/workflows: add a docs workflow

### LINKED ISSUE #111: Automated API documentation builds/publishing in CI
We don't currently do this locally because `pdoc3` doesn't support our minimum Python version (3.6), but there's no reason we can't do it in the CI and host the resulting HTML on a GitHub Pages site.

### LINKED ISSUE #130: Whittle down the public APIs
For the time being, we want to keep our publicly supported APIs to a bare minimum. In particular, that probably means just the following modules:

* `pip_audit.audit` (the `Auditor` class and `AuditOptions`)
* `pip_audit.service` (vulnerability services and interfaces)
* `pip_audit.dependency_source` (dependency sources and interfaces)

Everything else should be considered API private for the time being, and we should mark it as such. Additionally, some of the things in those public modules should also be marked private (like `resolvelib` helper classes/implementations).

Blocks #129.

woodruffw:
Closes #111.
Closes #130.


woodruffw:
Live here: https://trailofbits.github.io/pip-audit/

woodruffw:
Blocked on #130.

tetsuo-cpp (approved):
LGTM as far as the GitHub workflow is concerned.

.github/workflows/docs.yml
```diff
@@ -0,0 +1,28 @@
...
+      - uses: actions/setup-python@v2
+
+      - name: deps
+        run: make dev
+
+      - name: docs
+        run: make doc
+
+      - name: deploy
+        uses: peaceiris/actions-gh-pages@v3.8.0
```
di, line 23: Instead of using this action, can we just do the push ourselves instead? E.g. something like https://stackoverflow.com/questions/57921401/push-to-origin-from-github-action/58393457#58393457

woodruffw: Yeah, I'll give that a spin 🙂

woodruffw:
I've made every top-level module private, which makes our public API surface effectively zero for the time being.

We'll design a minimal public auditing API in a follow-up PR.

.github/workflows/docs.yml
woodruffw: Done; let me know what you think of the approach. It's a little bit more involved than the SO snippets, since we need to create a blank-slate "orphan" branch to avoid pushing up lots of incidental files.

.github/workflows/docs.yml
```diff
@@ -0,0 +1,38 @@
...
+      - name: docs
+        run: |
+          make doc
+          mv ./html/pip_audit /tmp/pip_audit
+
+      - name: deploy
+        run: |
+          # Configure git, checkout a new orphan branch for the docs.
+          git config --global user.name 'woodruffw'
+          git config --global user.email 'woodruffw@users.noreply.github.com'
```
di, line 28:
```suggestion
          git config --global user.name 'github.actions'
          git config --global user.email 'github-actions@github.com'
```
ref: https://github.com/actions/checkout#push-a-commit-using-the-built-in-token

di (approved):

PR #132, opened 2021-11-11
service/pypi: Use full cache path instead of directory name

### LINKED ISSUE #131: Unexpected behaviour when using the --cache-dir argument
## Bug description

When using `pip-audit` with the `--cache-dir` argument, it uses only the last component of the path passed in.  I believe this is because it is using the [name](https://docs.python.org/3/library/pathlib.html#pathlib.PurePath.name) property of the Path object https://github.com/trailofbits/pip-audit/blob/a2449e92dd2684173323407a8fb126ed77d1c5e7/pip_audit/service/pypi.py#L108

## Reproduction steps 

```shell
pip-audit --cache-dir /tmp/pip-audit/
```

You should observe that a `pip-audit` directory has been created in the current working directory rather than at `/tmp/pip-audit`

## Expected behavior

I would expect the cache to be at `/tmp/pip-audit/`

## Platform information

* OS name and version: Debian 11
* `pip-audit` version (`pip-audit -V`): 0.0.6
* Python version (`python -V` or `python3 -V`): 3.9.7
* `pip` version (`pip -V` or `pip3 -V`): 21.2.4

tetsuo-cpp:
Closes #131

### DIFF
--- a/test/service/test_pypi.py
+++ b/test/service/test_pypi.py
@@ -7,6 +7,7 @@ import requests
 from packaging.version import Version
 
 import pip_audit.service as service
+from pip_audit.service.pypi import _get_cache_dir
 
 cache_dir: Optional[tempfile.TemporaryDirectory] = None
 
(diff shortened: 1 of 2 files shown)


test/service/test_pypi.py
```diff
@@ -7,6 +7,7 @@
 from packaging.version import Version
 
 import pip_audit.service as service
+from pip_audit.service.pypi import _get_cache_dir
```
tetsuo-cpp, line 10: This is testing a private helper that isn't used in other modules. We tend not to do this, but I figured I should add some tests since I've already messed this up once.

If this isn't appropriate, feel free to revert the latest commit.

tetsuo-cpp:
Sorry about this... in my local testing I think I only ever set `--cache-dir` to a directory in the top level of my repo so I never noticed this behaviour. 🤦‍♂️

di (approved):

PR #145, opened 2021-11-30
pip_audit: Add skipped dependencies to audit summary

### LINKED ISSUE #141: Add skipped dependencies to audit summary

The `pip-audit` summary should give a full picture of what happened in the audit. At the moment, `pip-audit` logs warnings whenever it skips over a dependency (due to an invalid version, can't be found on PyPI, etc), however the summary doesn't contain this information. If `pip-audit` can't audit a particular dependency, this should be reflected in the summary/JSON output somehow.

There should be some indicator or section in the summary output containing the details of skipped dependencies.

tetsuo-cpp:
Closes #141

### DIFF
--- a/pip_audit/_format/columns.py
+++ b/pip_audit/_format/columns.py
@@ -3,7 +3,7 @@ Functionality for formatting vulnerability results as a set of human-readable co
 """
 
 from itertools import zip_longest
-from typing import Any, Dict, Iterable, List, Tuple
+from typing import Any, Dict, Iterable, List, Tuple, cast
 
 from packaging.version import Version
 
@@ -54,7 +54,7 @@ class ColumnsFormat(VulnerabilityFormat):
         for dep, vulns in result.items():
             if dep.is_skipped():
                 continue
-            assert isinstance(dep, service.ResolvedDependency)
+            dep = cast(service.ResolvedDependency, dep)
             for vuln in vulns:
                 vuln_data.append(self._format_vuln(dep, vuln))
 
@@ -77,7 +77,7 @@ class ColumnsFormat(VulnerabilityFormat):
         skip_data.append(skip_header)
         for dep, _ in result.items():
             if dep.is_skipped():
-                assert isinstance(dep, service.SkippedDependency)
+                dep = cast(service.SkippedDependency, dep)
                 skip_data.append(self._format_skipped_dep(dep))
 
         # If we only have the header, that means that we haven't skipped any dependencies
--- a/pip_audit/_format/columns.py
+++ b/pip_audit/_format/columns.py
@@ -52,6 +52,9 @@ class ColumnsFormat(VulnerabilityFormat):
             header.append("Description")
         vuln_data.append(header)
         for dep, vulns in result.items():
+            if dep.is_skipped():
+                continue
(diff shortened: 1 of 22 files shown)


tetsuo-cpp:
@woodruffw 
I liked the idea you described in https://github.com/trailofbits/pip-audit/issues/141#issuecomment-981785372 so I've modelled this PR after that.

pip_audit/_format/cyclonedx.py
```diff
@@ -21,6 +21,11 @@ def __init__(self, result: Dict[service.Dependency, List[service.VulnerabilityRe
         super().__init__()
 
         for (dep, vulns) in result.items():
+            # TODO(alex): Is there anything interesting we can do with skipped dependencies in
```
tetsuo-cpp, line 24: From looking at CycloneDX, I don't think there's a concept which captures this, so skipping is probably the right move.

test/service/test_pypi.py
```diff
@@ -22,8 +22,10 @@ def get(self, url, **kwargs):
 
 def test_pypi(cache_dir):
     pypi = service.PyPIService(cache_dir)
-    dep = service.Dependency("jinja2", Version("2.4.1"))
-    results: List[service.VulnerabilityResult] = dict(pypi.query_all([dep]))
+    dep = service.ResolvedDependency("jinja2", Version("2.4.1"))
+    results: Dict[service.Dependency, List[service.VulnerabilityResult]] = dict(
```
tetsuo-cpp, line 26: This isn't related to the change, I just noticed the type hints in this file are wrong. Is there a reason we don't run MyPy over the test code, or should I follow up with a change for that?

woodruffw: No good reason -- I was just lazy about only applying `mypy` to the library source. Yeah, let's do a follow-up PR for linting the tests as well.

pip_audit/_service/interface.py
```diff
@@ -11,10 +11,11 @@
 from packaging.version import Version
 
 
-@dataclass(frozen=True)
-class Dependency:
+# MyPy doesn't like abstract dataclasses so silence the checker here
+@dataclass(frozen=True)  # type: ignore[misc]
+class Dependency(ABC):
```
woodruffw, line 16: Let's de-abstract this and turn `is_skipped` into an instance check, so that MyPy can follow through the types here (and hopefully, if it's smart enough, figure out our subclass dispatch on `is_skipped` elsewhere).

I'll commit the change for that.

woodruffw: Looks like MyPy isn't smart enough to follow through the test here, unfortunately, probably because Python doesn't have a really sound way of enforcing strict subclassing. So I've added some runtime checks as unit tests instead.

pip_audit/_format/columns.py
```diff
@@ -52,6 +52,9 @@ def format(self, result: Dict[service.Dependency, List[service.VulnerabilityResu
             header.append("Description")
         vuln_data.append(header)
         for dep, vulns in result.items():
+            if dep.is_skipped():
+                continue
```
woodruffw, line 56: 👍

woodruffw (approved):
LGTM!

PR #148, opened 2021-12-01
pip, _cli: Add `--path` argument to mirror `pip list`

tetsuo-cpp:


pip_audit/_cli.py
```diff
@@ -227,10 +236,13 @@ def audit() -> None:
 
         source: DependencySource
         if args.requirements is not None:
+            if args.paths:
+                print("--requirement (-r) and --path arguments cannot be used together")
+                sys.exit(1)
```
woodruffw, line 241: We can probably use an exclusive argument group for this, to ensure at argument-parse-time that these are never used together: https://docs.python.org/3/library/argparse.html#argparse.ArgumentParser.add_mutually_exclusive_group

tetsuo-cpp: d1ecd3d

pip_audit/_dependency_source/pip.py
```diff
@@ -61,7 +65,9 @@ def collect(self) -> Iterator[Dependency]:
         # The `pip list` call that underlies `pip_api` could fail for myriad reasons.
         # We collect them all into a single well-defined error.
         try:
-            for (_, dist) in pip_api.installed_distributions(local=self._local).items():
+            for (_, dist) in pip_api.installed_distributions(
+                local=self._local, paths=list(self._paths)
```
tetsuo-cpp, line 69: @di 
I think making the API take `List[os.PathLike]` was a mistake. Because of `List` being invariant, we have to go through this contortion to pass type checking. 😬 

Sorry, I was testing within the `pip-api` tests and calling it through the REPL so I didn't notice the typing issue. Should I make a PR to make this a `Sequence[os.PathLike]` instead?

di: Sure!

woodruffw: I'll merge as-is, and we can handle that with another PR 🙂

woodruffw (approved):

westonsteimel:
Thanks so much!

woodruffw:
No problem! Please let us know if you run into any issues 🙂

PR #152, opened 2021-12-01
--desc flag needs argument on in README.md

JulianWgs:


woodruffw:
Thanks for catching that!

woodruffw (approved):

woodruffw:
#153 tweaks the behavior so that the original README example will be correct. Thanks again for bringing this to our attention.

PR #158, opened 2021-12-02
test, Makefile: Run MyPy across unit tests

### LINKED ISSUE #154: Run MyPy across unit tests
I spotted some incorrect type hints in some of our unit tests so I don't think we're running MyPy across our unit tests.

tetsuo-cpp:
Closes #154

### DIFF
--- a/test/dependency_source/test_resolvelib.py
+++ b/test/dependency_source/test_resolvelib.py
@@ -10,7 +10,7 @@ from resolvelib.resolvers import InconsistentCandidate, ResolutionImpossible
 
 from pip_audit._dependency_source import resolvelib
 from pip_audit._dependency_source.resolvelib import pypi_provider
-from pip_audit._service.interface import Dependency, ResolvedDependency
+from pip_audit._service.interface import ResolvedDependency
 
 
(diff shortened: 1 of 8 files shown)


test/dependency_source/test_resolvelib.py
```diff
@@ -49,7 +49,7 @@ def check_deps(resolved_deps: List[Dependency], expected_deps: List[Dependency])
 def test_resolvelib():
     resolver = resolvelib.ResolveLibResolver()
     req = Requirement("flask==2.0.1")
-    resolved_deps = dict(resolver.resolve_all([req]))
+    resolved_deps = dict(resolver.resolve_all(iter([req])))
```
woodruffw, line 52: Weird that all of these `iter(...)` calls are required, given that `List[...]` should already provide that type bound. I'll look into this.

woodruffw: Ah, this is probably because our various interfaces take and return `Iterator`s instead of `Iterable`s. We can selectively refine those another time.

woodruffw (approved):

PR #161, opened 2021-12-02
_service/pypi.py: Do not execute pip cache command when a custom directory is provided

JeffreyVdb:
In a scenario where the pip cache directory isn't available because the
user doesn't have permissions; pip will error with this message:

```
ERROR: pip cache commands cannot function since cache is disabled.
```

This currently cannot be mitigated by specifying the `--cache-dir` flag
because the `python -m pip cache dir` command is execute before this
predicate is evaluated. By using an assignment expression (PEP 572), we
can evaluate whether the expression evaluates to a truthy value in the
`elif` branch (saving it's return value) AFTER first evaluating the
custom directory value.

### DIFF
--- a/pip_audit/_service/pypi.py
+++ b/pip_audit/_service/pypi.py
@@ -109,12 +109,11 @@ def _get_pip_cache() -> str:
 
 
 def _get_cache_dir(custom_cache_dir: Optional[Path]) -> str:
-    pip_cache_dir: Optional[str] = (
-        _get_pip_cache() if _PIP_VERSION >= _MINIMUM_PIP_VERSION else None
-    )
     if custom_cache_dir is not None:
         return str(custom_cache_dir)
-    elif pip_cache_dir is not None:  # pragma: no cover
+    elif (
+        pip_cache_dir := _get_pip_cache() if _PIP_VERSION >= _MINIMUM_PIP_VERSION else None
+    ) is not None:  # pragma: no cover
         return pip_cache_dir
     else:
         fallback_path = os.path.join(Path.home(), ".pip-audit-cache")
--- a/pip_audit/_service/pypi.py
+++ b/pip_audit/_service/pypi.py
@@ -111,9 +111,9 @@ def _get_pip_cache() -> str:
 def _get_cache_dir(custom_cache_dir: Optional[Path]) -> str:
     if custom_cache_dir is not None:
         return str(custom_cache_dir)
-    elif (
-        pip_cache_dir := _get_pip_cache() if _PIP_VERSION >= _MINIMUM_PIP_VERSION else None
-    ) is not None:  # pragma: no cover
+
+    pip_cache_dir = _get_pip_cache() if _PIP_VERSION >= _MINIMUM_PIP_VERSION else None
+    if pip_cache_dir is not None:  # pragma: no cover
         return pip_cache_dir
     else:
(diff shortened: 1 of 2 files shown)


pip_audit/_service/pypi.py
```diff
@@ -109,12 +109,11 @@ def _get_pip_cache() -> str:
...
 
 def _get_cache_dir(custom_cache_dir: Optional[Path]) -> str:
-    pip_cache_dir: Optional[str] = (
-        _get_pip_cache() if _PIP_VERSION >= _MINIMUM_PIP_VERSION else None
-    )
     if custom_cache_dir is not None:
         return str(custom_cache_dir)
-    elif pip_cache_dir is not None:  # pragma: no cover
+    elif (
+        pip_cache_dir := _get_pip_cache() if _PIP_VERSION >= _MINIMUM_PIP_VERSION else None
```
tetsuo-cpp, line 115: We're trying to support Python 3.6 and later so we'll need to avoid the `:=`.

tetsuo-cpp (commented):
Great catch @JeffreyVdb! Thanks for this.

pip_audit/_service/pypi.py
JeffreyVdb: Fixed in: https://github.com/trailofbits/pip-audit/pull/161/commits/00ead12c134fb0315b4ef6e3026c81823ab0ab7b

pip_audit/_service/pypi.py
```diff
@@ -109,12 +109,11 @@ def _get_pip_cache() -> str:
...
 
 def _get_cache_dir(custom_cache_dir: Optional[Path]) -> str:
-    pip_cache_dir: Optional[str] = (
-        _get_pip_cache() if _PIP_VERSION >= _MINIMUM_PIP_VERSION else None
-    )
     if custom_cache_dir is not None:
         return str(custom_cache_dir)
-    elif pip_cache_dir is not None:  # pragma: no cover
+
+    pip_cache_dir = _get_pip_cache() if _PIP_VERSION >= _MINIMUM_PIP_VERSION else None
```
JeffreyVdb, line 115: Typing should be detected here to be `Optional[str]`, but I'm not sure if the convention is to explicitly write it as: 

```python
pip_cache_dir: Optional[str] = _get_pip_cache() if _PIP_VERSION >= _MINIMUM_PIP_VERSION else None
```

Let me know and I'll change it!

tetsuo-cpp (approved):
Nice! This looks good to me.

pip_audit/_service/pypi.py
woodruffw: Looks good as-is 🙂  -- we don't have a formal style preference for explicit/implicit types, but the general pattern so far has only been to use explicit types when `mypy` needs them.

woodruffw (approved):
LGTM as well. @JeffreyVdb, would you mind committing a CHANGELOG entry as well, under the "Fixed" heading for the upcoming release? You can use the already present entries as a style reference 🙂

JeffreyVdb:
> LGTM as well. @JeffreyVdb, would you mind committing a CHANGELOG entry as well, under the "Fixed" heading for the upcoming release? You can use the already present entries as a style reference 🙂

Rebased and added an entry to the `CHANGELOG`. Hopefully that's sufficient information.

di:
Thanks!

PR #178, opened 2021-12-03
pypi: add a retry loop for cache updates

### LINKED ISSUE #169: Frequent cache access error warnings on Win32
## Bug description

The cache writing code frequently runs into WinError 32 warnings due to parallel access by different processes. 
This might be due to the typical Windows Defender antivirus interfering with a create & rename operation. Files in use may not be renamed on windows.

## Reproduction steps 

Run pip-audit --local on a Windows machine with the default Defender Antivirus enabled.

## Expected behavior

No warnings, at least one retry.

## Screenshots and logs

```
DEBUG:pip_audit._cli:parsed arguments: Namespace(local=True, requirements=None, format=<OutputFormatChoice.Columns: 'columns'>, vulnerability_service=<VulnerabilityServiceChoice.Pypi: 'pypi'>, dry_run=False, strict=False, desc=<VulnerabilityDescriptionChoice.Auto: 'auto'>, cache_dir=WindowsPath('c:/temp/audit'), progress_spinner=<ProgressSpinnerChoice.On: 'on'>, timeout=15)
- Auditing zope.interface (5.4.0)
DEBUG:cachecontrol.controller:Looking up "https://pypi.org/pypi/alabaster/0.7.12/json" in the cache
DEBUG:cachecontrol.controller:No cache entry available
DEBUG:urllib3.connectionpool:Starting new HTTPS connection (1): pypi.org:443
DEBUG:urllib3.connectionpool:https://pypi.org:443 "GET /pypi/alabaster/0.7.12/json HTTP/1.1" 200 6108
DEBUG:cachecontrol.controller:Updating cache with response from "https://pypi.org/pypi/alabaster/0.7.12/json"
DEBUG:cachecontrol.controller:etag object cached for 1209600 seconds
DEBUG:cachecontrol.controller:Caching due to etag
WARNING:pip_audit._service.pypi:Failed to write to cache directory, performance may be degraded: [WinError 32] Der Prozess kann nicht auf die Datei zugreifen, da sie von einem anderen Prozess verwendet wird: 'c:\\temp\\audit\\8\\8\\c\\3\\c\\tmpfibz6hh_' -> 'c:\\temp\\audit\\8\\8\\c\\3\\c\\88c3c4038e9abf700f212a9b6a8b1294aca8991fc28f185b1fd2a782'
DEBUG:cachecontrol.controller:Looking up "https://pypi.org/pypi/argon2-cffi/21.1.0/json" in the cache
DEBUG:cachecontrol.controller:No cache entry available
```

## Platform information

* OS name and version:
* `pip-audit` version (`pip-audit -V`): pip-audit 1.0.1
* Python version (`python -V` or `python3 -V`): Python 3.9.7
* `pip` version (`pip -V` or `pip3 -V`): pip 21.3.1

## Additional context

Maybe using https://pypi.org/project/atomicwrites/ would help?

woodruffw:
Closes #169.

@schlenk or @matthewdeanmartin: could you give this a spin and see if it improves the caching situation on your hosts?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,7 +42,7 @@ All versions prior to 0.0.9 are untracked.
   the structure of source distributions was fixed
   ([#166](https://github.com/trailofbits/pip-audit/pull/166))
 
-* Dependency sources: a performance issue on Windows caused by cache failures
+* Vulnerability sources: a performance issue on Windows caused by cache failures
   was fixed ([#178](https://github.com/trailofbits/pip-audit/pull/178))
 
 ### Removed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,6 +42,9 @@ All versions prior to 0.0.9 are untracked.
   the structure of source distributions was fixed
   ([#166](https://github.com/trailofbits/pip-audit/pull/166))
 
+* Dependency sources: a performance issue on Windows caused by cache failures
+  was fixed ([#178](https://github.com/trailofbits/pip-audit/pull/178))
+
 ### Removed
 
 ## [1.0.1] - 2021-12-02
--- a/pip_audit/_service/pypi.py
+++ b/pip_audit/_service/pypi.py
@@ -7,7 +7,6 @@ import logging
 import os
 import subprocess
 import sys
-import time
 from pathlib import Path
 from subprocess import run
 from tempfile import NamedTemporaryFile
@@ -72,7 +71,7 @@ class _SafeFileCache(FileCache):
         except (IOError, OSError):  # pragma: no cover
             pass
 
-        # We don't want to use lock files since `pip` isn't going to recognise those. We should
+        # We don't want to use lock files since `pip` isn't going to recognize those. We should
         # write to the cache in a similar way to how `pip` does it. We create a temporary file,
         # then atomically replace the actual cache key's filename with it. This ensures
         # that other concurrent `pip` or `pip-audit` instances don't read partial data.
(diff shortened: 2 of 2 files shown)


schlenk:
Gave it a try, but did not solve the issue. I took a closer look with ProcessMonitor, to see which syscalls happend and the virus scanner does open the files, but is not really interfering in this case.

The actual issue why the writes failed is this:
https://stackoverflow.com/questions/7147577/programmatically-rename-open-file-on-windows
(see also https://devblogs.microsoft.com/oldnewthing/20211022-00/?p=105822 )

You try to rename the file while INSIDE the `with NamedTemporaryFile()` block, so the file handle is still open at the time the rename/replace() happens. This leads to a SHARING_VIOLATION and the replace() fails.

NamedTemporaryFile with the O_TEMPORARY flag actually sets SHARE_DELETE, but that would not help, as it deletes the file on close, which is not what you really want here. And python does not provide a simple way to open a file with share mode delete ( you can do it with a bit of ctypes, CreateFile, os.openfd() & mscvrt.open_osfhandle(), but it is a bit clumsy.). 

So one way that works (on windows) is to put the os.replace() outside the with block, so it happens after the file is closed.

matthewdeanmartin:
>  And python does not provide a simple way to open a file with share mode delete ( you can do it with a bit of ctypes, CreateFile, os.openfd() & mscvrt.open_osfhandle(), but it is a bit clumsy.).
> 

While hacking on an a tangentially related problem and I ran into that same temporary file bug on windows and stack overflow suggest this exact solution so that python temp files would work on windows, especially if someone is trying to do something fancy with them before the close:

https://github.com/matthewdeanmartin/caniuseonlywheels/blob/master/test/custom_temp_file.py

woodruffw:
Thanks for digging into that! Yeah, that makes more sense and should be much simpler to test. I'll update the PR accordingly.

woodruffw:
Okay, this should be good for another round of tests. Could one of you give it a spin again?

schlenk:
Works for me now. No more warnings and the cache gets populated and is reused on a second run.

woodruffw:
Awesome!

di (approved):

PR #185, opened 2021-12-05
_cli: Create --verbose flag

### LINKED ISSUE #177: Can we have a --verbose flag for diagnosing performance problems?

`pip-audit` will finish in a few seconds but `pip-audit -r requirement.txt` hangs forever. If I interrupt it, sometimes it is in the middle of trying to upgrade pip, sometimes it is in the middle of trying to POST something via requests (as inferred from the call stack at point of killing the process)

I'd like `pip-audit` to not hang on windows 10, when using python 3.9, pipenv or pipx, when on corporate VPN, with shell of git-bash or cmd.exe... but I can't send my laptop to y'all to diagnose what's up so I don't really expect a bug report to get any attention.

So I'd like a --verbose flag so I can diagnose what is happening and where the code is hanging (is the company's net-nanny banning the domain of something because it is uncategorized, who knows? is it hitting a free service and getting rate limited? Is it calling `safety` without an API key like pipenv used to?)

Forking the project, but that takes a lot of time.

**Additional context**

None

tetsuo-cpp:
Closes #177

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -32,7 +32,7 @@ A clear and concise description of what you expected to happen.
 
 If applicable, add screenshots to help explain your problem.
 
-Similarly, if applicable and possible, re-run the command with `PIP_AUDIT_LOGLEVEL=debug` exported,
+Similarly, if applicable and possible, re-run the command with `--verbose`,
 and paste the logs in the code block below:
 
 ```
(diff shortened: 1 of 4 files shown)


tetsuo-cpp:
The default logging level is `INFO` and the only level more verbose than that is `DEBUG`. So I didn't really see a way to make the `--verbose` flag additive like it is with `pip`.

woodruffw (approved):

PR #192, opened 2021-12-06
Refactor `AuditState`

### LINKED ISSUE #191: Refactor `AuditState`
`AuditState` is our nebulous blob-object that we pass throughout the codebase to keep the CLI responsive. Right now it's passed as an `Optional[AuditState]` everywhere, to provide for the eventual use case where a non-CLI consumer doesn't need it.

As a result, every use site looks roughly like this:

```python
if state is not None:
    state.update_state(...)
```

This is a lot of boilerplate. Instead, we should turn `AuditState` into a proxy object for a list of state-receiving objects. Then, each callsite can skip the `None` check.

woodruffw:
Saves us a bunch of `Optional`s, and simplifies the API a bit. Also gives us the flexibility to reuse `AuditState` for things other than spinners.

No public-facing changes.

Closes #191.


di (approved):

tetsuo-cpp (approved):
LGTM! This is nicer.

PR #205, opened 2021-12-13
setup: use CacheControl[filecache]

### LINKED ISSUE #203: pip-audit has unhealthy dependencies
## Bug description

Running [Deadpendency](https://deadpendency.com) on pip-audit results in three unhealthy pip-audit (direct/indirect) dependencies:

- lockfile – 📦
  - ❌ Last package release over 24 months ago. Last release: 2015-11-25
  - ⚠ A source repository was not identified.
- requirements-parser – 📖 📦
  - ❌ No commits within the last 18 months. Last commit: 2020-03-31
  - ❌ Last package release over 24 months ago. Last release: 2018-01-12
- webencodings – 📖 📦
  - ❌ No commits within the last 18 months.
  - ❌ Last package release over 24 months ago. Last release: 2017-04-05

## Reproduction steps 

[Install Deadpendency](https://deadpendency.com/docs/getting-started#installation) on a public GitHub repo that has pip-audit in its requirements.txt. 

An example Deadpendency report (including the dependencies listed above) can be found here: https://github.com/ICTU/quality-time/pull/3021/checks?check_run_id=4501775840

## Expected behavior

pip-audit shouldn't rely on unhealthy dependencies.

## Platform information

* OS name and version: n/a
* `pip-audit` version (`pip-audit -V`): pip-audit 1.1.1
* Python version (`python -V` or `python3 -V`): Python 3.10.0
* `pip` version (`pip -V` or `pip3 -V`): pip 21.3.1

woodruffw:
~~Clarifies why we need `lockfile` even though it doesn't appear anywhere in `pip-audit`'s own code.~~

With this, all of the dependencies in #203 are accounted for on our side.

Closes #203.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,7 +3,7 @@ pip-audit
 
 <!--- BADGES: START --->
 ![CI](https://github.com/trailofbits/pip-audit/workflows/CI/badge.svg)
-[![PyPI version](https://badge.fury.io/py/pip-audit.svg)](https://badge.fury.io/py/pip-audit)
+[![PyPI version](https://badge.fury.io/py/pip-audit.svg)](https://pypi.org/project/pip-audit)
 [![Packaging status](https://repology.org/badge/tiny-repos/python:pip-audit.svg)](https://repology.org/project/python:pip-audit/versions)
 <!--- BADGES: END --->
 
--- a/setup.py
+++ b/setup.py
@@ -36,6 +36,7 @@ setup(
         "resolvelib>=0.8.0",
         "html5lib>=1.1",
         "CacheControl>=0.12.10",
+        # NOTE: Non-default internal dependency of CacheControl.
         "lockfile>=0.12.2",
         "cyclonedx-python-lib>=0.11.1",
     ],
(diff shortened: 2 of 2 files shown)


di:
Do we even need to specify this in `setup.py` if it's a transitive dependency? Looks like it's only installed as part of [the `cachecontrol[filecache]` extra](https://github.com/ionrock/cachecontrol/blob/3bd694551d28445723c18f4ad8c44bce55ff9b5f/setup.py#L25), should we be specifying that instead?

woodruffw:
> Do we even need to specify this in `setup.py` if it's a transitive dependency? Looks like it's only installed as part of [the `cachecontrol[filecache]` extra](https://github.com/ionrock/cachecontrol/blob/3bd694551d28445723c18f4ad8c44bce55ff9b5f/setup.py#L25), should we be specifying that instead?

Ah yeah, we should just do that instead! Thanks for catching that.

tetsuo-cpp (approved):
LGTM!

PR #212, opened 2022-01-12
pip, _fix: Implement `--fix` for `PipSource`

tetsuo-cpp:


tetsuo-cpp:
There's a bit of cleanup and UX to figure out like how to notify users what upgrades took place. But this is the general idea I'm going for. It upgrades Flask 0.5 to Flask 1.0 successfully.

The main deviation from what we discussed earlier is that instead of doing `pip install --upgrade`, I have some code to "resolve" the fix version. I do that by looking at the fix versions of all the vulnerabilities, choosing the latest one (that includes all the fixes), then checking it against the vulnerability service. If there are vulnerabilities in that version, I repeat until I either find a clean version with no vulnerabilities or I find a vulnerability with no fix version (in which case I'll fail the `--fix`).

My reasoning behind this is that the latest version of a package can have a reported vulnerability so blindly doing `pip install --upgrade` doesn't seem to be the right thing to do given that we have fix version information at our disposal. Also generally speaking, the more versions we jump, the more likely the user is going to experience API breakages. So I think we should upgrade the minimum distance possible without introducing new known vulnerabilities.

pip_audit/_dependency_source/pip.py
```diff
@@ -87,6 +89,18 @@ def collect(self) -> Iterator[Dependency]:
...
+        """
+        Fixes a dependency version in this `PipSource`.
+        """
+        fix_cmd = [sys.executable, "-m", "pip", "install", f"{dep.name}=={fix_version}"]
+        try:
+            subprocess.run(
+                fix_cmd, check=True, stdout=subprocess.DEVNULL, stderr=subprocess.DEVNULL
+            )
+        except subprocess.CalledProcessError as cpe:
+            raise RuntimeError from cpe
```
tetsuo-cpp, line 102: Need to handle this properly.

pip_audit/_fix.py
```diff
@@ -0,0 +1,52 @@
...
+    # prospective resolved fix version
+    current_version = dep.version
+    current_vulns = vulns
+    while current_vulns:
+
+        def get_earliest_fix_version(fix_versions: List[Version]) -> Version:
+            for v in fix_versions:
+                if v > current_version:
+                    return v
+            raise RuntimeError
```
tetsuo-cpp, line 45: And this.

woodruffw:
> My reasoning behind this is that the latest version of a package can have a reported vulnerability so blindly doing pip install --upgrade doesn't seem to be the right thing to do given that we have fix version information at our disposal. Also generally speaking, the more versions we jump, the more likely the user is going to experience API breakages. So I think we should upgrade the minimum distance possible without introducing new known vulnerabilities.

This reasoning makes sense to me, and the underlying algorithm makes sense to me as well!

pip_audit/_dependency_source/interface.py
```diff
@@ -26,6 +27,20 @@ def collect(self) -> Iterator[Dependency]:  # pragma: no cover
...
+        """
+        raise NotImplementedError
+
+    def fix_all(self, fix_req: Iterator[Tuple[Dependency, Version]]) -> None:
+        """
+        Upgrade a collection of dependencies to their associated fix versions.
+        """
+        for (dep, fix_version) in fix_req:
+            self.fix(dep, fix_version)
+
```
woodruffw, line 43: Good call adding these to the abstract interface!

pip_audit/_fix.py
woodruffw: Something like `FixResolutionImpossible`? That would mirror the `ResolutionImpossible` that `resolvelib` provides.

woodruffw:
Overall structure looks good to me! In terms of notifying users, I was thinking that we'd amend our normal output.

Without `--fix`:

```
$ pip-audtit
Found 12 known vulnerabilities in 18 packages
<output depending on --format>
```

With `--fix`:

```
$ pip-audit --fix
Found 12 known vulnerabilities in 18 packages and fixed 9 vulnerabilities in 4 packages
<output depending on --format>
```

In particular, both our columnar and JSON formats should support emitting a summary of the fixes performed (and those skipped, including the reason the fix failed).

di:
I agree we should apply the "minimum" fix here.

Let's think about exit codes as well: should `pip-audit --fix` fail if it can't fix all vulnerabilities? IMO yes.

tetsuo-cpp:
> In particular, both our columnar and JSON formats should support emitting a summary of the fixes performed (and those skipped, including the reason the fix failed).

> How does that sound?

Yep, that's a good plan. In the interest of not letting this PR accumulate too much code, I'll make separate issues for these and follow up.

pip_audit/_cli.py
```diff
@@ -280,11 +287,34 @@ def audit() -> None:
...
+        for fix_version in resolve_fix_versions(service, result):
+            if not fix_version.is_skipped():
+                fix_version = cast(ResolvedFixVersion, fix_version)
+                try:
+                    source.fix(fix_version)
+                    fixed_pkg_count += 1
+                    fixed_vuln_count += len(result[fix_version.dep])
+                except DependencySourceError as dse:
+                    fix_version = SkippedFixVersion(fix_version.dep, str(dse))
+            fixes.append(fix_version)
```
tetsuo-cpp, line 305: These will eventually get passed into the `VulnerabilityFormat` interface.

pip_audit/_fix.py
```diff
@@ -0,0 +1,92 @@
...
+
+from pip_audit._service import (
+    Dependency,
+    ResolvedDependency,
+    VulnerabilityResult,
+    VulnerabilityService,
+)
+
+
+@dataclass(frozen=True)
```
tetsuo-cpp, line 18: I've followed a similar pattern to how we handle dependencies if you recall that @woodruffw.

tetsuo-cpp:
The CI breakage is due to #213

pip_audit/_dependency_source/pip.py
```diff
@@ -87,6 +90,27 @@ def collect(self) -> Iterator[Dependency]:
...
+            "pip",
+            "install",
+            f"{fix_version.dep.canonical_name}=={fix_version.version}",
+        ]
+        try:
+            subprocess.run(
+                fix_cmd, check=True, stdout=subprocess.DEVNULL, stderr=subprocess.DEVNULL
+            )
+        except subprocess.CalledProcessError as cpe:
+            raise PipSourceError(
```
woodruffw, line 109: Nitpick: we might want to consider creating a specialized `DependencyFixError` that each dependency source in turn specializes.

So the hierarchy would be: `DependencySourceError -> DependencyFixError -> PipFixError | RequirementFixError`.

pip_audit/_fix.py
woodruffw: Yep, this LGTM!

pip_audit/_cli.py
woodruffw: SGTM!

xref #215.

woodruffw (approved):

PR #221, opened 2022-01-14
Handle plurals correctly in output

di:
Fixing a small nit. The test is kind of an experiment and thus a bit rough as it's not super easy to test this right now, ideally this will become cleaner in time.


tetsuo-cpp (approved):
LGTM!

PR #225, opened 2022-01-19
requirement: Implement `--fix` for `RequirementSource`

### LINKED ISSUE #214: Implement `--fix` for `RequirementSource`
After merging #212, we should mirror this functionality in `RequirementSource` and edit the underlying requirements file.

tetsuo-cpp:
Closes #214

### DIFF
--- a/pip_audit/_dependency_source/requirement.py
+++ b/pip_audit/_dependency_source/requirement.py
@@ -6,6 +6,7 @@ from pathlib import Path
 from typing import Iterator, List, Set, cast
 
 from packaging.requirements import Requirement
+from packaging.specifiers import SpecifierSet
 from pip_api import parse_requirements
 from pip_api.exceptions import PipError
 
@@ -79,11 +80,33 @@ class RequirementSource(DependencySource):
             except DependencyResolverError as dre:
                 raise RequirementSourceError("dependency resolver raised an error") from dre
 
-    def fix(self, fix_version: ResolvedFixVersion) -> None:  # pragma: no cover
+    def fix(self, fix_version: ResolvedFixVersion) -> None:
         """
         Fixes a dependency version for this `RequirementSource`.
         """
-        raise NotImplementedError
+        for filename in self.filenames:
+            self.state.update_state(
+                f"Fixing dependency {fix_version.dep.name} ({fix_version.dep.version} => "
+                f"{fix_version.version})"
+            )
+            self._fix_file(filename, fix_version)
+
+    def _fix_file(self, filename: Path, fix_version: ResolvedFixVersion) -> None:
+        try:
+            reqs = parse_requirements(filename=filename)
+        except PipError as pe:
+            raise RequirementSourceError("requirement parsing raised an error") from pe
+
+        # Invoke the dependency resolver to turn requirements into dependencies
+        req_values: List[Requirement] = [Requirement(str(req)) for req in reqs.values()]
+
+        with open("fixed.txt", "w+") as f:
+            for req in req_values:
+                if req.name == fix_version.dep.name and req.specifier.contains(
+                    fix_version.dep.version
+                ):
+                    req.specifier = SpecifierSet(f"=={fix_version.version}")
+                f.write(str(req))
 
 
 class RequirementSourceError(DependencySourceError):
--- a/pip_audit/_dependency_source/requirement.py
+++ b/pip_audit/_dependency_source/requirement.py
@@ -3,11 +3,12 @@ Collect dependencies from one or more `requirements.txt`-formatted files.
 """
 
 from pathlib import Path
-from typing import Iterator, List, Set, cast
+from typing import Iterator, List, Set, Union, cast
 
 from packaging.requirements import Requirement
 from packaging.specifiers import SpecifierSet
 from pip_api import parse_requirements
+from pip_api._parse_requirements import UnparsedRequirement
 from pip_api.exceptions import PipError
 
 from pip_audit._dependency_source import (
@@ -92,18 +93,34 @@ class RequirementSource(DependencySource):
             self._fix_file(filename, fix_version)
 
     def _fix_file(self, filename: Path, fix_version: ResolvedFixVersion) -> None:
+        # TODO(alex): Preserve comments
+        #
+        # Reparse the requirements file. We want to rewrite each line to the new requirements file
+        # and only modify the lines that we're fixing.
         try:
-            reqs = parse_requirements(filename=filename)
+            reqs = parse_requirements(filename=filename, include_invalid=True)
         except PipError as pe:
             raise RequirementSourceError("requirement parsing raised an error") from pe
 
-        # Invoke the dependency resolver to turn requirements into dependencies
-        req_values: List[Requirement] = [Requirement(str(req)) for req in reqs.values()]
+        # Convert requirements types from pip-api's vendored types to our own
+        req_values: List[Union[Requirement, UnparsedRequirement]] = list()
+        for value in reqs.values():
+            if isinstance(value, UnparsedRequirement):
+                req_values.append(value)
+            else:
+                req_values.append(Requirement(str(value)))
 
+        # TODO(alex): Toggle what file to write to
+        #
+        # Now write out the new requirements file
         with open("fixed.txt", "w+") as f:
             for req in req_values:
-                if req.name == fix_version.dep.name and req.specifier.contains(
-                    fix_version.dep.version
+                if (
+                    isinstance(req, Requirement)
+                    and req.name == fix_version.dep.name
+                    and req.specifier.contains(fix_version.dep.version)
+                    and not req.specifier.contains(fix_version.version)
+                    and (req.marker is None or req.marker.evaluate())
                 ):
                     req.specifier = SpecifierSet(f"=={fix_version.version}")
                 f.write(str(req))
--- a/pip_audit/_dependency_source/requirement.py
+++ b/pip_audit/_dependency_source/requirement.py
@@ -135,10 +135,10 @@ class RequirementSource(DependencySource):
                     req.name == fix_version.dep.name
                     and req.specifier.contains(fix_version.dep.version)
                     and not req.specifier.contains(fix_version.version)
-                    and (req.marker is None or req.marker.evaluate())
                 ):
                     req.specifier = SpecifierSet(f"=={fix_version.version}")
-                f.write(str(req))
+                assert req.marker is None or req.marker.evaluate()
+                f.write(str(req) + os.linesep)
 
     def _recover_files(self, tmp_files: List[IO[str]]) -> None:
         for (filename, tmp_file) in zip(self.filenames, tmp_files):
--- a/pip_audit/_dependency_source/requirement.py
+++ b/pip_audit/_dependency_source/requirement.py
@@ -129,7 +129,7 @@ class RequirementSource(DependencySource):
         req_list: List[Requirement] = [Requirement(str(req)) for req in reqs.values()]
 
         # Now write out the new requirements file
-        with open(filename, "w") as f:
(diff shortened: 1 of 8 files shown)


tetsuo-cpp:
@woodruffw @di 
I wanted to run this by you both before going too far with this. At the moment, I use `pip_api.parse_requirements` and then rewrite the file line by line, replacing requirements that correspond to the fixed dependency.

This could be optimised to do all the fixes in-memory and then write the file out at the end but it doesn't fit the `fix` interface and I don't anticipate this being a serious performance issue, so I'm thinking we can do it like this initially.

* One problem with this approach is that we lose comments in the requirement file since they don't get parsed by `parse_requirements`. I'm thinking that the solution might be to add a `keep_comments` optional argument to `pip_api.parse_requirements` that returns strings for comments also.

* The other issue is that the fact that we support multiple requirements files makes this harder to get right (apply fixes transactionally). With a single file, this is easy as we just need to construct the new file with the fixes and then we can move it into place to replace the old one. Multiple files throw a spanner in the works since if one of our moves fails, we'll have some requirement files with fixes applied and some without, so we'll have to keep the old ones around and then have some recovery procedure to undo the moves. If we want to keep things simple and avoid dealing with this, we could also just say that `--fix` only supports a single `-r` argument.

What are your thoughts on these two points?

woodruffw:
> This could be optimised to do all the fixes in-memory and then write the file out at the end but it doesn't fit the `fix` interface and I don't anticipate this being a serious performance issue, so I'm thinking we can do it like this initially.

Yeah, this sounds good to me. Our execution is probably overwhelmingly dominated by network round-trips and `pip` subprocesses, so this isn't something we need to immediately optimize.

> * One problem with this approach is that we lose comments in the requirement file since they don't get parsed by `parse_requirements`. I'm thinking that the solution might be to add a `keep_comments` optional argument to `pip_api.parse_requirements` that returns strings for comments also.

@di should probably have the final say on this, but IMO we should skip on comment handling for the MVP. I *do* think we should ultimately preserve comments, but we might need to rethink the `parse_requirements` API to do so, since we'll need to support both freestanding and inline comments:

```
# foo
blah==1.2.3 # bar
```

> * The other issue is that the fact that we support multiple requirements files makes this harder to get right (apply fixes transactionally).

Hmm, this is a good point. Maybe this suggests that we redesign our initial `pip_api._fix` APIs to be more transactional to begin with, since we have a similar "partial fix" problem even in non-`-r` mode: we can imagine that upgrading `foo 1.2.3 -> 1.2.4` is a correct fix, but that `pip install -U foo==1.2.4` fails for whatever reason. If that happens amidst other successful fixes, we perhaps should fail the entire `--fix` (and revert any upgrades already performed).

If we go the above route, we'll probably want to refactor `DependencySource.fix` into a non-abstract method like this:

```python
def fix(self, fix_versions: Iterator[ResolvedFixVersion]):
    with self.fix_transaction():
        for fix in fix_versions:
            self.fix_one(fix)
```

...where `fix_one` is the previous body of `DependencySource.fix` for each implementation, and `fix_transaction` is another abstract method that provides a transactional context manager.

OTOH, that's a lot of complexity, and maybe `--fix` should really just be "buyer beware": there's no guarantee that the `--fix` will succeed or leave multiple requirements inputs in a consistent state, only that failures *are* accompanied by error handling on our side (a nonzero exit). With that approach, it's the user's responsibility to rollback any changes made.

tetsuo-cpp:
> Hmm, this is a good point. Maybe this suggests that we redesign our initial pip_api._fix APIs to be more transactional to begin with, since we have a similar "partial fix" problem even in non--r mode: we can imagine that upgrading foo 1.2.3 -> 1.2.4 is a correct fix, but that pip install -U foo==1.2.4 fails for whatever reason. If that happens amidst other successful fixes, we perhaps should fail the entire --fix (and revert any upgrades already performed).

Personally, I think this is ok. Our summary specifically states that we fixed X vulnerabilities out of Y packages (compared with the Z vulnerabilities found in the audit), which indicates that we will fix everything we can but we don't guarantee we can fix everything. At the moment, if the package update fails, we yield a `SkippedFixVersion` and `pip-audit` behaves similarly to if we can't resolve a fix version.

But I don't think it's ok to partially apply a single fix to some requirements files but not others.

tetsuo-cpp:
How does this look?

I've attempted to deal with the case where to fail to write to one of the requirements files (or some other error gets thrown). I start by copying all requirements files to temporary files, then attempt to rewrite the requirements files. If we're successful, the temporary files will automatically get cleaned up. If not, I restore the requirements files by copying the temporary files back to the original location.

pip_audit/_dependency_source/requirement.py
```diff
@@ -79,14 +87,77 @@ def collect(self) -> Iterator[Dependency]:
...
         """
-        raise NotImplementedError
+        with ExitStack() as stack:
+            # Make temporary copies of the existing requirements files. If anything goes wrong, we
+            # want to copy them back into place and undo any partial application of the fix.
+            tmp_files: List[IO[str]] = [
+                stack.enter_context(TemporaryFile(mode="w")) for _ in self.filenames
+            ]
+            for (filename, tmp_file) in zip(self.filenames, tmp_files):
+                with open(filename, "r") as f:
```
woodruffw, line 101: Nitpick: Use the `pathlib` APIs whenever we have a `Path` object:

```suggestion
                with filename.open("r") as f:
```

pip_audit/_dependency_source/requirement.py
```diff
@@ -79,14 +87,77 @@ def collect(self) -> Iterator[Dependency]:
...
+                    and not req.specifier.contains(fix_version.version)
+                    and (req.marker is None or req.marker.evaluate())
+                ):
+                    req.specifier = SpecifierSet(f"=={fix_version.version}")
+                f.write(str(req))
+
+    def _recover_files(self, tmp_files: List[IO[str]]) -> None:
+        for (filename, tmp_file) in zip(self.filenames, tmp_files):
+            try:
+                with open(filename, "w") as f:
```
woodruffw, line 145:
```suggestion
                with filename.open("w") as f:
```

pip_audit/_dependency_source/requirement.py
```diff
@@ -79,14 +87,77 @@ def collect(self) -> Iterator[Dependency]:
...
+                    shutil.copyfileobj(tmp_file, f)
+            except Exception as e:
+                # Not much we can do at this point since we're already handling an exception. Just
+                # log the error and try to recover the rest of the files.
+                logger.warning(f"encountered an exception during file recovery: {e}")
+                continue
```
woodruffw, line 151: We might want to consider using `rename()` or `replace()` for this, since it's guaranteed to be atomic and those calls have fewer error conditions than `copyfileobj`.

tetsuo-cpp: 256cf25

tetsuo-cpp:
I'm still working on getting this tested. Mocking out things so the tests don't mess around with the filesystem has been a bit of a pain.

> This overall structure looks good to me! This approach also looks good, and I think it's what we'll want to encapsulate with the "fix transaction" interfaces above. But that can be a task for a follow-up PR 🙂

Hmm, I'm not following. The way I see it, the "fix transaction" interface is only necessary if we're intending to roll back fixes for other vulnerabilities in the case that one of them failed. Is that what we want? The recovery scheme I'm using here is for when a single vulnerability exists in multiple requirements files and we didn't manage to fix it in all files (as opposed to rolling back successful fixes for vulnerabilities across all files because a vulnerability further down the list couldn't be fixed).

test/dependency_source/test_requirement.py
```diff
@@ -94,3 +96,57 @@ def test_requirement_source_duplicate_dependencies(monkeypatch):
...
+        return self.mock_file
+
+    def __str__(self) -> str:
+        return self.path
+
+    def __fspath__(self) -> str:
+        return self.path
+
+
+def test_requirement_source_fix(monkeypatch):
```
tetsuo-cpp, line 132: This was already troublesome enough, but I'll need to figure out a way to mock out the recovery parts.

Instead of continuing down this path, I'm thinking I might just have a helper in `conftest.py` to create a real requirements file in the form of a temporary file and inject text into it. That way, I don't need to mock anything.

woodruffw:
> The way I see it, the "fix transaction" interface is only necessary if we're intending to roll back fixes for other vulnerabilities in the case that one of them failed. Is that what we want? 

Whoops, this was a misunderstanding on my part. Yeah, we don't need the "transaction" interface in this case.

test/dependency_source/test_requirement.py
woodruffw: Yeah, having a real requirements file as a test fixture seems like a more reliable/less painful approach. If we have to mock anything at all for these tests, we should probably limit it to just the fix directives and not the I/O or FS ops.

woodruffw: (i.e., the test should be "pass a tempfile with some contents in, and assert that those contents have changed as specified by a phony fix directive")

test/dependency_source/test_requirement.py
```diff
@@ -94,3 +98,186 @@ def test_requirement_source_duplicate_dependencies(monkeypatch):
...
+        [req_file, other_req_file],
+        [
+            ResolvedFixVersion(
+                dep=ResolvedDependency(name="flask", version=Version("0.5")), version=Version("1.0")
+            )
+        ],
+    )
+
+
+def test_requirement_source_fix_marker(req_file, other_req_file):
```
tetsuo-cpp, line 179: This test and the one directly below document a few shortcomings of the current approach. Any requirement with markers that don't apply to the current environment are lost. Also, comments are lost as we discussed before.

woodruffw: Thanks for adding negative tests for these. We should also make sure to add corresponding tracking issues on the `pip-api` side of things.

test/conftest.py
```diff
@@ -59,3 +61,17 @@ def cache_dir():
...
+    req_file = tempfile.NamedTemporaryFile(delete=False)
+    yield Path(req_file.name)
+    os.remove(req_file.name)
+
+
+@pytest.fixture
+def other_req_file():
+    req_file = tempfile.NamedTemporaryFile(delete=False)
+    yield Path(req_file.name)
+    os.remove(req_file.name)
```
woodruffw, line 77: N.B.: You can avoid two fixtures here by making this into a higher-order fixture, e.g.:

```python
@pytest.fixture
def req_file():
    def _req_file():
        # original body here
    return _req_file
```

I'll do that as part of merge fixups.

woodruffw (approved):
LGTM overall! I left some small nits that I can take care of.

test/dependency_source/test_requirement.py
```diff
@@ -94,3 +98,186 @@ def test_requirement_source_duplicate_dependencies(monkeypatch):
...
+        ],
+    )
+
+
+def test_requirement_source_fix_comments(req_file, other_req_file):
+    # `pip-api` automatically filters out comments
+    _check_fixes(
+        [
+            "# comment here\nflask==0.5\n",
+            "requests==1.0\n# another comment\nflask==0.5",
```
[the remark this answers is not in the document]
woodruffw: Oh, I understand what's going on -- each of these is an entirely separate requirements file. We can probably clarify that a bit with docstring blocks.

test/dependency_source/test_requirement.py
woodruffw: https://github.com/di/pip-api/issues/120

di:
> I'm still working on getting this tested. Mocking out things so the tests don't mess around with the filesystem has been a bit of a pain.

There might be some useful fixtures for test isolation in https://github.com/di/pip-api/blob/dec3a5e30c911b794763483ed985960a6732a40e/tests/conftest.py you could use to improve this.

PR #228, opened 2022-01-31
Make the CLI more responsive

### LINKED ISSUE #227: Make `pip-audit`'s spinner still more responsive
This is not triaged/prioritized since it's just a braindump.

#56 and #89 made `pip-audit` more responsive, but we can still do better.

In particular, the spinner still pauses when `pip-audit` is performing any kind of subprocess (i.e., calling out to `pip` internally).

We can probably improve this by adding some kind of synchronous monitoring of each subprocess -- we can either update the spinner naively (i.e., on a cadence, once every `X` milliseconds while waiting for the subprocess to terminate), or by monitoring the process's output and updating the spinner e.g. each time we see a new line on `stdout` or `stderr`.

Upsides:

* The `pip-audit` CLI becomes even more responsive, particularly in subprocess-heavy modes like `-r` and `--fix`

Downsides:

* We probably have to use one of the messier `subprocess` APIs, like `subprocess.communicate`

woodruffw:
Closes #227.

Demo:

[asciicast](https://asciinema.org/a/465724)

### DIFF
--- a/pip_audit/_subprocess.py
+++ b/pip_audit/_subprocess.py
@@ -1,0 +1,54 @@
+"""
+A thin `subprocess` wrapper for making long-running subprocesses more
+responsive from the `pip-audit` CLI.
+"""
+
+import os.path
+import shlex
+import subprocess
+from subprocess import Popen
+from typing import Sequence
+
+from ._state import AuditState
+
+
+class CalledProcessError(Exception):
+    """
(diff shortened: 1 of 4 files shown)


woodruffw:
The main remaining bottleneck in the CLI's spinner is virtual environment creation itself, i.e. `EnvBuilder.create()`. We can probably hook further into that with the public interface it provides, but I figured that's best kept for a follow-up.

PR #229, opened 2022-02-01
requirement, pypi: Add a --require-hashes flag

### LINKED ISSUE #172: Add a --require-hashes flag
Similar to `pip`'s [`--require-hashes` flag](https://pip.pypa.io/en/stable/cli/pip_install/#cmdoption-require-hashes), when auditing requirements files, `pip-audit` should support a `--require-hashes` flag that:

- requires all auditable dependencies to be pinned and hashed
- disables dependency resolution

Additionally, it would likely be desirable for `pip-audit` to ensure that the hashes that are present match hashes provided via PyPI's JSON API.

tetsuo-cpp:
Closes #172

- [x] Bump `pip-api` version
- [x] Needs: https://github.com/di/pip-api/pull/126

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,7 +71,7 @@ python -m pip_audit --help
 usage: pip-audit [-h] [-V] [-l] [-r REQUIREMENTS] [-f FORMAT] [-s SERVICE]
                  [-d] [-S] [--desc [{on,off,auto}]] [--cache-dir CACHE_DIR]
                  [--progress-spinner {on,off}] [--timeout TIMEOUT]
-                 [--path PATHS] [-v] [--fix]
+                 [--path PATHS] [-v] [--fix] [--require-hashes]
 
 audit the Python environment for dependencies with known vulnerabilities
 
@@ -115,6 +115,10 @@ optional arguments:
                         setting it to `debug` (default: False)
   --fix                 automatically upgrade dependencies with known
                         vulnerabilities (default: False)
+  --require-hashes      require a hash to check each requirement against, for
+                        repeatable audits; this option is implied when any
+                        package in a requirements file has a --hash option.
+                        (default: False)
 ```
 <!-- @end-pip-audit-help@ -->
 
--- a/README.md
+++ b/README.md
@@ -117,7 +117,7 @@ optional arguments:
                         vulnerabilities (default: False)
   --require-hashes      require a hash to check each requirement against, for
                         repeatable audits; this option is implied when any
-                        package in a requirements file has a --hash option.
+                        package in a requirements file has a `--hash` option.
                         (default: False)
 ```
 <!-- @end-pip-audit-help@ -->
--- a/pip_audit/_cli.py
+++ b/pip_audit/_cli.py
@@ -239,6 +239,12 @@ def _parser() -> argparse.ArgumentParser:
         action="store_true",
         help="automatically upgrade dependencies with known vulnerabilities",
     )
+    parser.add_argument(
+        "--require-hashes",
+        action="store_true",
+        help="require a hash to check each requirement against, for repeatable audits; this option "
+        "is implied when any package in a requirements file has a --hash option.",
+    )
     return parser
 
 
@@ -272,7 +278,10 @@ def audit() -> None:
         if args.requirements is not None:
             req_files: List[Path] = [Path(req.name) for req in args.requirements]
             source = RequirementSource(
-                req_files, ResolveLibResolver(args.timeout, args.cache_dir, state), state
+                req_files,
+                ResolveLibResolver(args.timeout, args.cache_dir, state),
+                args.require_hashes,
+                state,
             )
         else:
             source = PipSource(local=args.local, paths=args.paths, state=state)
--- a/pip_audit/_cli.py
+++ b/pip_audit/_cli.py
@@ -243,7 +243,7 @@ def _parser() -> argparse.ArgumentParser:
         "--require-hashes",
         action="store_true",
         help="require a hash to check each requirement against, for repeatable audits; this option "
-        "is implied when any package in a requirements file has a --hash option.",
+        "is implied when any package in a requirements file has a `--hash` option.",
     )
     return parser
 
--- a/pip_audit/_cli.py
+++ b/pip_audit/_cli.py
@@ -268,6 +268,10 @@ def audit() -> None:
     output_desc = args.desc.to_bool(args.format)
     formatter = args.format.to_format(output_desc)
 
+    # The `--require-hashes` flag is only valid with requirements files
+    if args.require_hashes and args.requirements is None:
+        parser.error("The --require-hashes flag can only be used with --requirement (-r)")
+
     with ExitStack() as stack:
         actors = []
         if args.progress_spinner:
--- a/pip_audit/_service/interface.py
+++ b/pip_audit/_service/interface.py
@@ -4,8 +4,8 @@ of vulnerability information for fully resolved Python packages.
 """
 
 from abc import ABC, abstractmethod
-from dataclasses import dataclass
-from typing import Iterator, List, Optional, Tuple
+from dataclasses import dataclass, field
+from typing import Dict, Iterator, List, Tuple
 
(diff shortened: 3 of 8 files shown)


tetsuo-cpp:
This needs cleanup + testing but I wanted to engage early on this one since the intended behaviour isn't straightforward.

@di You mentioned that the `--require-hashes` flag should skip dependency resolution. Did you mean that we shouldn't resolve dependencies for each requirement and instead, only audit what is explicitly listed in the requirements file?

I was also wondering how hashes should work when the `--require-hashes` flag isn't supplied. My thinking is that when the `--require-hashes` flag is supplied, every requirement in the file needs to have a hash. But if it isn't supplied, we should still check any hashes that we find against PyPI.

woodruffw:
> You mentioned that the `--require-hashes` flag should skip dependency resolution. Did you mean that we shouldn't resolve transitive dependencies for each requirement and instead, only audit what is explicitly listed in the requirements file?
> 
> I was also wondering how hashes should work when the `--require-hashes` flag isn't supplied. My thinking is that when the `--require-hashes` flag is supplied, every requirement in the file needs to have a hash AND we check them against PyPI. But if the flag isn't supplied, we should still check any hashes that we find against PyPI and fail if it doesn't match.

This is an indirect answer to both: we can skip dependency resolution entirely, since `pip` stipulates the following:

1. If *any* dependency has a content hash, then all dependencies have content hashes
2. (Transitively) All dependencies must be *explicitly* listed in the requirements file if hashing is enabled

In other words: our requirements parser in `pip-api` should preserve those properties: we should raise an error if we're given a requirements file with dependencies that are missing hashes, and we should *not* do dependency resolution because only the present hashed dependencies should be considered.

(Looking at things more, I'm not actually sure that we *need* the `--require-hashes` flag -- `pip` stipulates that passing it is identical to any requirement having hashes attached to it. But I guess we should preserve it as a no-op, since that's what `pip` seems to do?

pip_audit/_service/interface.py
```diff
@@ -54,6 +54,7 @@ class ResolvedDependency(Dependency):
     """
 
     version: Version
+    hash: Optional[str] = None
```
woodruffw, line 57: This should probably be a `Dict[str, List[str]]` instead, since every requirement is allowed to have multiple hashes, even multiple hashes of the same kind (since `foo==1.2.3` might have both an `sdist` and `bdist_wheel` distribution).

tetsuo-cpp: Makes sense. I also need to modify the logic to check all of the hashes.

tetsuo-cpp:
> In other words: our requirements parser in `pip-api` should preserve those properties: we should raise an error if we're given a requirements file with dependencies that are missing hashes, and we should _not_ do dependency resolution because only the present hashed dependencies should be considered.

Ah! Right, I misunderstood how that works. Thanks for explaining.

> (Looking at things more, I'm not actually sure that we need the --require-hashes flag -- pip stipulates that passing it is identical to any requirement having hashes attached to it. But I guess we should preserve it as a no-op, since that's what pip seems to do?

I think it's still important because your file may not have ANY hashes in it. If you pass in `--require-hashes`, it will complain whereas it otherwise would succeed.

woodruffw:
> I think it's still important because your file may not have ANY hashes in it. If you pass in `--require-hashes`, it will complain whereas it otherwise would succeed.

Ah! You're absolutely right. Yes, we need it for that case.

di:
> @di You mentioned that the --require-hashes flag should skip dependency resolution. Did you mean that we shouldn't resolve transitive dependencies for each requirement and instead, only audit what is explicitly listed in the requirements file?

Thinking about this more, I think we still need to do dependency resolution, and ensure that we have hashes for all subdependencies, because this is what `pip` does as well:

```
$ cat req.txt
sampleproject==2.0.0 \
    --hash=sha256:2b0c55537193b792098977fdb62f0acbaeb2c3cfc56d0e24ccab775201462e04 \
    --hash=sha256:d99de34ffae5515db43916ec47380d3c603e9dead526f96581b48c070cc816d3

$ python -m pip install -r req.txt
Collecting sampleproject==2.0.0 (from -r req.txt (line 1))
  Using cached https://files.pythonhosted.org/packages/b8/f7/dd9223b39f683690c30f759c876df0944815e47b588cb517e4b9e652bcf7/sampleproject-2.0.0-py3-none-any.whl
Collecting peppercorn (from sampleproject==2.0.0->-r req.txt (line 1))
ERROR: In --require-hashes mode, all requirements must have their versions pinned with ==. These do not:
    peppercorn from https://files.pythonhosted.org/packages/14/84/d8d9c3f17bda2b6f49406982546d6f6bc0fa188a43d4e3ba9169a457ee04/peppercorn-0.6-py3-none-any.whl#sha256=46125cad688a9cf3b08e463bcb797891ee73ece93602a8ea6f14e40d1042d454 (from sampleproject==2.0.0->-r req.txt (line 1))
```

> I think it's still important because your file may not have ANY hashes in it. If you pass in `--require-hashes`, it will complain whereas it otherwise would succeed.

Agreed!

tetsuo-cpp:
> Thinking about this more, I think we still need to do dependency resolution, and ensure that we have hashes for all subdependencies, because this is what pip does as well:

Can we leave that validation to `pip` and not attempt to do that validation ourselves? I imagine that's going to introduce a significant chunk of code reinventing what `pip` does which will have to be removed during integration, since `pip` already contains this logic.

pip_audit/_cli.py
```diff
@@ -239,6 +239,12 @@ def _parser() -> argparse.ArgumentParser:
         action="store_true",
         help="automatically upgrade dependencies with known vulnerabilities",
     )
+    parser.add_argument(
+        "--require-hashes",
+        action="store_true",
+        help="require a hash to check each requirement against, for repeatable audits; this option "
+        "is implied when any package in a requirements file has a --hash option.",
```
woodruffw, line 246:
```suggestion
        "is implied when any package in a requirements file has a `--hash` option.",
```

README.md
```diff
@@ -115,6 +115,10 @@ optional arguments:
                         setting it to `debug` (default: False)
   --fix                 automatically upgrade dependencies with known
                         vulnerabilities (default: False)
+  --require-hashes      require a hash to check each requirement against, for
+                        repeatable audits; this option is implied when any
+                        package in a requirements file has a --hash option.
```
woodruffw, line 120:
```suggestion
                        package in a requirements file has a `--hash` option.
```

pip_audit/_cli.py
woodruffw: We should also exit with an error if someone tries to use `--require-hashes` **without** `-r`. An argument group is probably the right approach for that.

woodruffw:
> Can we leave that validation to pip and not attempt to do it ourselves? I imagine that's going to introduce a significant chunk of code reinventing what pip does which will have to be removed during integration, since pip already contains this logic.

I'm inclined to agree with this, although it does deviate our handling of `--require-hashes` from that of `pip` (our `--require-hashes` is more like `--no-deps --require-hashes`). My reasoning is that `pip-audit` isn't in the business of telling the user whether `pip install ...` would succeed, and that the overwhelming majority of hashed requirements files should be valid already (since, if they're like Warehouse's, they're checked in after successful local and CI usage).

Thoughts @di?

pip_audit/_cli.py
tetsuo-cpp: [f667aea](https://github.com/trailofbits/pip-audit/pull/229/commits/f667aea6939823e9c9b5a741d9642b9bd245cc15)

I don't believe I can use an argument group to say that an argument requires something else. The `-r` flag is also part of a mutually exclusive group with `--local`.

Instead I've just written a check that invokes `parser.error` if we use `--require-hashes` without `-r`.

woodruffw (approved):
Overall structure LGTM here!

di:
I suppose the `pip install` with an incompletely hashed requirements file, or a hashed requirements file that's missing dependencies, would fail to install anyways, so the user would be protected regardless.

di (approved):

PR #238, opened 2022-02-09
_cli, resolvelib: Support for custom indices

### LINKED ISSUE #46: Support for custom indices
For the time being, our MVP is scoped to just support for PyPI. However, it's worth considering what we'd require in order to support custom package indices (whether PyPI mirrors or entirely separate private indexes).

Some potential points of issue:

* Our #23 adaptor will need to be sufficiently generic, adapting either PyPI or another index under the hood as configured.
* We *might* need custom index support from `pip-api`? This is unclear, since our only use of `pip-api` in these contexts would be for requirements parsing, and the requirements file *itself* shouldn't make any references to the index.

tetsuo-cpp:
Closes #46

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,14 @@ All versions prior to 0.0.9 are untracked.
   conjunction with `-r` to check that all requirements in the file have an
   associated hash ([#229](https://github.com/trailofbits/pip-audit/pull/229))
 
+* CLI: The `--index-url` flag has been added, allowing users to use custom
+  package indices when running with the `-r` flag
+  ([#238](https://github.com/trailofbits/pip-audit/pull/238))
+
+* CLI: The `--extra-index-url` flag has been added, allowing users to use
+  multiple package indices when running with the `-r` flag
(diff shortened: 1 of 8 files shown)


tetsuo-cpp:
I'll spin up a DevPi instance tomorrow to test this out more thoroughly.

woodruffw:
Overall structure LGTM! I agree about the additional testing.

di:
> I'll spin up a DevPi instance tomorrow to test this out more thoroughly.

You might be able to test with https://test.pypi.org instead, could be easier.

tetsuo-cpp:
> > I'll spin up a DevPi instance tomorrow to test this out more thoroughly.
> 
> You might be able to test with https://test.pypi.org instead, could be easier.

Thanks! I did some testing with the test PyPI.

If my requirement simply says `pip`, it audits 22.0.3 from prod PyPI since it is a later version than anything on the test PyPI. If I constrain to `pip==18.1`, I can get it to audit the test PyPI version.

di (approved):

woodruffw (approved):
Awesome, LGTM!

PR #246, opened 2022-03-03
Support `pyproject.toml`

### LINKED ISSUE #83: Support `pyproject.toml`
Breakout from #73: We should support dependency tools that store their information in `pyproject.toml`.

tetsuo-cpp:
Closes #83

- [x] Unit tests

### DIFF
--- a/pip_audit/_cli.py
+++ b/pip_audit/_cli.py
@@ -161,10 +161,7 @@ def _parser() -> argparse.ArgumentParser:
         help="audit the given requirements file; this option can be used multiple times",
     )
     dep_source_args.add_argument(
-        "-p",
-        "--pyproject",
-        type=argparse.FileType("r"),
-        help="audit the given `pyproject.toml` file",
+        "project_path", type=Path, nargs="?", help="audit a local Python project at the given path"
     )
     parser.add_argument(
         "-f",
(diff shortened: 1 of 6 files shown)


pip_audit/_cli.py
```diff
@@ -159,6 +160,12 @@ def _parser() -> argparse.ArgumentParser:
         dest="requirements",
         help="audit the given requirements file; this option can be used multiple times",
     )
+    dep_source_args.add_argument(
+        "-p",
+        "--pyproject",
+        type=argparse.FileType("r"),
+        help="audit the given `pyproject.toml` file",
+    )
```
di, line 168: See my comment here: https://github.com/trailofbits/pip-audit/issues/81#issuecomment-1032988316

We shouldn't need a `-p` or `--pyproject` flag, we should auto-detect the presence of a `pyproject.toml` file when auditing a local source tree.

di (commented):
Took a quick pass, I think this looks good except for the CLI. Let's make sure to add documentation w/ examples as well.

pip_audit/_cli.py
woodruffw: Agreed; `pip-audit .` should intelligently select either `setup.py` (not implemented yet) or `pyproject.toml`, with a preference for the latter if it has dependencies listed.

tetsuo-cpp: My bad, I missed that comment. Should be an easy fix though.

pip_audit/_cli.py
```diff
@@ -317,13 +323,12 @@ def audit() -> None:
...
-                pyproject_file,
-                ResolveLibResolver(index_urls, args.timeout, args.cache_dir, state),
-                state,
-            )
+        elif args.project_path is not None:
+            # Determine which kind of project file exists in the project path
+            _source = _create_project_source(args.project_path, state)
+            if _source is None:
+                parser.error(f"couldn't find project file at path: {args.project_path}")
+            source = _source
```
woodruffw, line 340: IMO it's okay to reduce this a bit -- we can have `_create_project_source` exit the program instead of returning `None` as a signal. I'll take care of that while shepherding this.

woodruffw (approved):
LGTM!

PR #255, opened 2022-04-30
cli, dependency_source: support `--no-deps`

### LINKED ISSUE #168: scan without resolving/downloading dependencies
maybe i'm misunderstanding, but it seems that `pip-audit` cannot scan the contents of a requirements file in the simplest way imaginable: for each entry, check if the specified version has known vulnerabilities.

currently, running against a requirements file with exclusively exact version matches (`foo==x.y`), like this

```
$ pip-audit -r requirements.txt 
```

… downloads and (temporarily) installs various packages, which not only takes forever, but also eventually fails because some packages cannot be found – they come from a private index server, which is a different issue.

as a work-around, installing everything (also downloads and installs) and then using `pip-audit --local` seems to work, but this is not a convenient solution.

i understand that in the general case, requirements files may not be exhaustive and may not have exact versions, in which case running a resolver etc makes total sense, but that's not always the case.

some background: the use case is that my requirements files are produced by `pip-tools` and hence constitute the exact list of pinned versions used by my application, and i would like to integrate `pip-audit` into a continuous integration job. i also run `pip-compile` (from `pip-tools`) in that job to warn about potential package upgrades, some of which come from private index servers, and it would be great if `pip-audit` can be added to it.

woodruffw:
This adds `pip-audit --no-deps`, allowing a user to explicitly skip the dependency resolution step when auditing a requirements source (e.g. `-r requirement.txt`). This functions similarly to the extant `--require-hashes` mode in terms of disabling dependency resolution, albeit without actually enforcing a hash policy.

Needs docs and unit tests.

Closes #168.


pip_audit/_cli.py
```diff
@@ -326,6 +332,22 @@ def audit() -> None:
...
             parser.error("The --extra-index-url flag can only be used with --requirement (-r)")
+        elif args.no_deps:
+            parser.error("The --no-deps flag can only be used with --requirement (-r)")
+
+    # Nudge users to consider alternate workflows.
+    if args.require_hashes and args.no_deps:
+        logger.warning("The --no-deps flag is redundant when used with --require-hashes")
+
+    if args.no_deps:
+        logger.warning(
```
tetsuo-cpp, line 343: Nice! I like that.

tetsuo-cpp (approved):
LGTM!

di (approved):

PR #259, opened 2022-05-01
Add that pip-aduit does not work in pre-commit.ci

cooperlees:
- pre-commit.ci blocks network connections / pip
- So share this knowledge and config to skip it into README so others don't get suprised like I did
Test: https://github.com/pypa/bandersnatch/pull/1116

### DIFF
--- a/README.md
+++ b/README.md
@@ -314,6 +314,11 @@ arguments in your pre-commit config. An example config using requirements file c
     hooks:
       -   id: pip-audit
           args: ["-r", "requirements.txt"]
+
+ci:
+  # Leave pip-audit to only run locally and not in CI
+  # pre-commit.ci does not allow network calls
+  skip: [pip-audit]
 ```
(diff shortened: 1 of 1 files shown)


woodruffw (approved):

woodruffw:
Thanks! That's too bad (but completely understandable) about the network access.

cooperlees:
Ya, totally agree with pre-commit.ci's decision here too. But it can work locally / in your own action if you wish. So i feel it does not harm allow people to choose to use it if they want. I'm going a separate GHA route. I only planned to use pre-commit as I thought it was the hotness ...

woodruffw:
Yep. If it helps, I believe the folks at Santander have a prebuilt action here: https://github.com/Santandersecurityresearch/pip-audit-action

(I believe their action only supports `pip-audit -r ...` at the moment, but that might be ideal for your use case!)

PR #267, opened 2022-05-02
Makefile: add `make dev`; refine env rebuilding

woodruffw:
Now that we have `setup.cfg`, our Makefile should also track it.


woodruffw:
Blocked on #268.

PR #275, opened 2022-05-10
cli: allow ignoring specific vulnerability IDs

### LINKED ISSUE #245: Explicitly ignore certain PYSECs

Certain libraries have unfixed PYSECs that are acknowledged yet not a problem in certain scenarios.

For example, there is [PYSEC-2017-49](https://github.com/pypa/advisory-database/blob/main/vulns/kerberos/PYSEC-2017-49.yaml) for the kerberos library, however the function impacted by said PYSEC is already [documented](https://github.com/apple/ccs-pykerberos/blob/master/pysrc/kerberos.py#L52-L54) to be dangerous by the lib maintainers, and essentially if your application doesn't use said function it's not impacted.

A flag like `--ignore` and/or a config file wherein PYSECs to be ignored can be defined, akin to how pylint and flake8 do it, this would allow people to more easily integrate the tool with CI/CD pipelines as any detected vulnerability currently triggers a non-zero exit-code

None

**Additional context**

None

Elkasitu:
This commit introduces a new flag to pip-audit's CLI (--ignore-vuln)
that takes a string representing a vulnerability id, if any of the
packages scanned detect said vulnerability, it will be ignored in the
final report.

This allows users of pip-audit to ignore certain vulnerabilities that
they may deem not exploitable for whatever reason.

The flag can be used multiple times to ignore multiple vulnerabilities.

Closes #245

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,14 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 All versions prior to 0.0.9 are untracked.
 
+## [Unreleased]
+
+### Added
+
+* CLI: The `--ignore-vuln` option has been added, allowing users to
+  specify vulnerability IDs to ignore during the final report.
+  ([#275](https://github.com/trailofbits/pip-audit/pull/275))
+
 ## [2.2.1] - 2022-05-02
 
 ### Fixed
--- a/pip_audit/_cli.py
+++ b/pip_audit/_cli.py
@@ -287,6 +287,14 @@ def _parser() -> argparse.ArgumentParser:
         # argparse's default renderer uses __repr__ and produces
         # a pretty unpleasant help message.
     )
+    parser.add_argument(
+        "--ignore-vuln",
+        type=str,
+        action="append",
+        dest="ignore_vulns",
+        default=[],
+        help="ignore a specific vulnerability by its vulnerability ID",
+    )
     return parser
 
 
@@ -390,6 +398,8 @@ def audit() -> None:
         pkg_count = 0
         vuln_count = 0
         skip_count = 0
+        vuln_ignore_count = 0
+        vulns_to_ignore = set(args.ignore_vulns)
         for (spec, vulns) in auditor.audit(source):
             if spec.is_skipped():
                 spec = cast(SkippedDependency, spec)
@@ -401,6 +411,10 @@ def audit() -> None:
             else:
                 spec = cast(ResolvedDependency, spec)
                 state.update_state(f"Auditing {spec.name} ({spec.version})")
+            if vulns_to_ignore:
+                filtered_vulns = [v for v in vulns if not v.has_any_id(vulns_to_ignore)]
+                vuln_ignore_count += len(vulns) - len(filtered_vulns)
+                vulns = filtered_vulns
             result[spec] = vulns
             if len(vulns) > 0:
                 pkg_count += 1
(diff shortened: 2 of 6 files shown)


CHANGELOG.md
```diff
@@ -6,6 +6,14 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 All versions prior to 0.0.9 are untracked.
 
+## [Unreleased]
+
+### Added
+
+* CLI: The `--ignore-vuln` option has been added, allowing users to
+  specify PYSEC ids to ignore during the final report.
```
woodruffw, line 14: Nit: We support IDs other than PYSEC, so we should probably just say "vulnerability IDs"

pip_audit/_cli.py
```diff
@@ -401,6 +410,10 @@ def audit() -> None:
             else:
                 spec = cast(ResolvedDependency, spec)
                 state.update_state(f"Auditing {spec.name} ({spec.version})")
+            if args.ignore_vuln:
+                filtered_vulns = [v for v in vulns if v.id not in args.ignore_vuln]
```
woodruffw, line 414: Nit: We also probably need to check the each result's `aliases` set here, since the ID might be a non-PYSEC ID (or the vuln might not have a PYSEC ID at all).

That's a bit messy to do inline, so we might want to consider a new `VulnerabilityResult.has_id` API.

Elkasitu: Ok, so if I understand correctly this `has_id` (or `has_ids`) method should take a list of IDs (PYSEC, CVE, etc.) and check whether the input list intersects with the `VulnerabilityResult`'s id + aliases? and if so return `True`, `False` otherwise?

woodruffw: Yep, correct. I guess we need to check the intersection for all ignored IDs, so maybe `has_any_id(set)`?

Elkasitu: I added this as a separate commit, let me know if you wish to have it squashed

pip_audit/_cli.py
```diff
@@ -401,6 +410,10 @@ def audit() -> None:
             else:
                 spec = cast(ResolvedDependency, spec)
                 state.update_state(f"Auditing {spec.name} ({spec.version})")
+            if args.ignore_vulns:
+                filtered_vulns = [v for v in vulns if not v.has_any_id(set(args.ignore_vulns))]
```
woodruffw, line 414: Tiny nit, but let's maybe memoize `set(args.ignore_vulns)` somewhere outside of the loop? That way we're not reallocating it on each audit step.

pip_audit/_cli.py
```diff
@@ -287,6 +287,14 @@ def _parser() -> argparse.ArgumentParser:
         # argparse's default renderer uses __repr__ and produces
         # a pretty unpleasant help message.
     )
+    parser.add_argument(
+        "--ignore-vuln",
+        type=str,
+        action="append",
+        dest="ignore_vulns",
+        default=[],
+        help="ignore a specific vulnerability by its PYSEC id",
```
di, line 296: This is no longer just PYSEC IDs, correct? (Also, please capitalize 'ID' here)

woodruffw: Yeah, this should just be `by its ID`.

woodruffw (approved):
LGTM, I'll give @di another chance to review but otherwise I can merge tonight 🙂

di (approved):

woodruffw:
Thanks a ton @Elkasitu!

PR #283, opened 2022-05-24
cli, state: replace `progress` spinner with `rich.status`

### LINKED ISSUE #282: Replace `progress` with `rich`
`pip` semi-recently switched to `rich` for its progress bars/spinners. We should do the same, since:

1. It simplifies the dependency graph for vendoring `pip-audit` into `pip`
2. `rich` is nicer anyways, and has easier to use APIs

woodruffw:
This simplifies the spinner implementation quite a bit, and brings us slightly closer in line with `pip`'s already vendored dependencies.

Closes #282.

### DIFF
--- a/pip_audit/_state.py
+++ b/pip_audit/_state.py
@@ -4,13 +4,11 @@ well as a progress spinner implementation for use with CLI applications.
 """
 
 import logging
-import os
 from abc import ABC, abstractmethod
 from logging.handlers import MemoryHandler
-from typing import Any, Dict, List, Sequence
+from typing import Any, List, Sequence
 
-from progress import SHOW_CURSOR
-from progress.spinner import Spinner as BaseSpinner
+from rich.console import Console
 
 
 class AuditState:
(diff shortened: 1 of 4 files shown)


woodruffw:
Demos:

Old:

[asciicast](https://asciinema.org/a/496954)

New:

[asciicast](https://asciinema.org/a/496953)

Note that the new one has a faster startup, and is generally smoother.

pip_audit/_state.py
```diff
@@ -99,24 +97,20 @@ def finalize(self) -> None:
...
-
-        Any remaining keyword arguments are forwarded onto the constructor of the underlying
-        `BaseSpinner` implementation.
+        Initialize the `AuditSpinner`.
         """
 
-        super().__init__(message=message, **kwargs)
+        self._console = Console()
+        # NOTE: audits can be quite fast, so we need a pretty high refresh rate here.
+        self._spinner = self._console.status(message, spinner="line", refresh_per_second=30)
```
woodruffw, line 113: N.B.: If we wanted to be extra fancy, we could allow users to configure the `spinner=...` here, rather than using the same "line" spinner that `progress` uses.

Rich has a decent collection of spinners to choose from: 

https://github.com/Textualize/rich/blob/aa7926c1431eebfb2ccaab9f3b63a4ac6cd8dfe6/rich/_spinners.py#L22

di (approved):

PR #284, opened 2022-05-24
osv: fix incorrect schema assumptions

woodruffw:
We shouldn't assume that any fields other than `id`
or `modified` exist, and we currently don't check the latter.

### DIFF
--- a/pip_audit/_service/osv.py
+++ b/pip_audit/_service/osv.py
@@ -89,9 +89,7 @@ class OsvService(VulnerabilityService):
             if description is None:
                 description = "N/A"
 
-            aliases = set()
-            if "aliases" in vuln:
-                aliases.update(vuln["aliases"])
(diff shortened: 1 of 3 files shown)


pip_audit/_service/osv.py
```diff
@@ -77,10 +80,28 @@ def query(self, spec: Dependency) -> Tuple[Dependency, List[VulnerabilityResult]
...
+            # details, if present. However, neither is required.
+            description = vuln.get("summary")
+            if description is None:
+                description = vuln.get("details")
+            if description is None:
+                description = "N/A"
+
+            aliases = set()
+            if "aliases" in vuln:
+                aliases.update(vuln["aliases"])
```
di, line 94:
```suggestion
            aliases = set(vuln.get("aliases", []))
```

di (approved):

Issue #286, opened 2022-05-25
--index-url Argument not recognized


msachsenhauser:
I'm trying to use pip-audit with a corporate package-proxy using the --index-url argument stated in --help .

```
poetry export -f requirements.txt --output requirements.txt
poetry run pip-audit -r requirements.txt --index-url https://some.internal.url/pypi/pypi-virtual/simple --skip-editable --desc on
# the use of poetry doesn't make any difference
```

Unfortunatelly, it seems like pip-audit ignores --index-url and tries to access pypi.org directly (_what is blocked by the corporate proxy_). The provided package-index-address isn't mentioned in the protocol below.

> Skipping virtualenv creation, as specified in config file.
> Traceback (most recent call last):
>   File "C:\path\to\venv\lib\site-packages\urllib3\connection.py", line 174, in _new_conn
>     conn = connection.create_connection(
>   File "C:\path\to\venv\lib\site-packages\urllib3\util\connection.py", line 95, in create_connection
>     raise err
>   File "C:\path\to\venv\lib\site-packages\urllib3\util\connection.py", line 85, in create_connection
>     sock.connect(sa)
> socket.timeout: timed out
> 
> During handling of the above exception, another exception occurred:
> 
> Traceback (most recent call last):
>   File "C:\path\to\venv\lib\site-packages\urllib3\connectionpool.py", line 703, in urlopen
>     httplib_response = self._make_request(
>   File "C:\path\to\venv\lib\site-packages\urllib3\connectionpool.py", line 386, in _make_request
>     self._validate_conn(conn)
>   File "C:\path\to\venv\lib\site-packages\urllib3\connectionpool.py", line 1040, in _validate_conn
>     conn.connect()
>   File "C:\path\to\venv\lib\site-packages\urllib3\connection.py", line 358, in connect
>     conn = self._new_conn()
>   File "C:\path\to\venv\lib\site-packages\urllib3\connection.py", line 179, in _new_conn
>     raise ConnectTimeoutError(
> urllib3.exceptions.ConnectTimeoutError: (<urllib3.connection.HTTPSConnection object at 0x000001DB1C111550>, 'Connection to pypi.org timed out. (connect timeout=15)')
> 
> During handling of the above exception, another exception occurred:
> 
> Traceback (most recent call last):
>   File "C:\path\to\venv\lib\site-packages\requests\adapters.py", line 440, in send
>     resp = conn.urlopen(
[...]

PR #287, opened 2022-05-25
cli, service: better error handling for connection errors

woodruffw:
These can occur if the user is behind a corporate firewall
or other network policy that blocks access to PyPI,
or in the unlikely event of a PyPI API outage.

See #286.


woodruffw:
For example, here's the error message we produce when we can't connect to PyPI's JSON API:

```console
(env) work:pip-audit william$ pip-audit .
ERROR:pip_audit._cli:Could not connect to PyPI's vulnerability feed
ERROR:pip_audit._cli:Tip: your network may be blocking this service. Try another service with `-s SERVICE`
```

di (approved):

PR #297, opened 2022-06-14
requirement, test: Correct `--fix` for subdependencies in requirements files

### LINKED ISSUE #291: The `--fix` flag doesn't work for dependencies of top-level requirements
I'm still working on a repro for this but just from reading the code, I'm pretty sure that this isn't correct. When running `--fix` on the current environment it works like this:

- Figure out what version to upgrade the broken dependency to.
- Upgrade it and ignore any other package that uses that dependency. If we have packages that don't work properly with the upgraded dependency, too bad.

I think this is reasonable for an automated fix.
However, for requirements files our approach doesn't work well. We're currently doing:

- Figure out what version to upgrade the broken dependency to.
- Scan the requirements file and look for a matching dependency and update it.

However, transitive dependencies won't necessarily appear in the requirements file unless the requirements have pinned hashes. So in this case, the fix won't be applied to the requirements file. I think we have a few options here:

1. Rewrite the `--fix` logic to work differently for requirements files. We'll need something that identifies which top-level requirement a dependency belongs to and then upgrade the top-level requirement, check whether it fixed the vulnerability, rinse, repeat.
2. Make `--fix` write a new line to the requirements file explicitly listing the fixed dependency if it isn't already there.
3. Only allow `--fix` in combination with `--require-hashes` or `--no-deps` since we know we'll have the full set of dependencies in the requirements file.

tetsuo-cpp:
Closes #291

- [x] Figure out test failure on 3.10
- [x] CHANGELOG

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,11 @@ All versions prior to 0.0.9 are untracked.
   `pyproject.toml` dependency source
   ([#300](https://github.com/trailofbits/pip-audit/pull/300))
 
+* Vulnerability fixing: the `--fix` flag now works for vulnerabilities found in
+  requirement subdependencies. A new line is now added to the requirement file
+  to explicitly pin the offending subdependency
+  ([#297](https://github.com/trailofbits/pip-audit/pull/297))
+
 ## [2.3.2] - 2022-05-14
 
 ### Changed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,13 @@ All versions prior to 0.0.9 are untracked.
 
 ## [Unreleased]
 
+### Fixed
+
+* Vulnerability fixing: the `--fix` flag now works for vulnerabilities found in
+  requirement subdependencies. A new line is now added to the requirement file
+  to explicitly pin the offending subdependency
+  ([#297](https://github.com/trailofbits/pip-audit/pull/297))
+
 ## [2.3.3]
 
 ### Changed
@@ -23,11 +30,6 @@ All versions prior to 0.0.9 are untracked.
   dependency resolver behavior now work correctly when auditing a
   `pyproject.toml` dependency source
   ([#300](https://github.com/trailofbits/pip-audit/pull/300))
-
-* Vulnerability fixing: the `--fix` flag now works for vulnerabilities found in
-  requirement subdependencies. A new line is now added to the requirement file
-  to explicitly pin the offending subdependency
-  ([#297](https://github.com/trailofbits/pip-audit/pull/297))
 
 ## [2.3.2] - 2022-05-14
 
--- a/pip_audit/_dependency_source/requirement.py
+++ b/pip_audit/_dependency_source/requirement.py
@@ -9,7 +9,7 @@ import shutil
 from contextlib import ExitStack
 from pathlib import Path
 from tempfile import NamedTemporaryFile
-from typing import IO, Dict, Iterator, List, Set, Union, cast
+from typing import IO, Dict, Iterator, List, Set, Tuple, Union, cast
 
 from packaging.requirements import Requirement
 from packaging.specifiers import SpecifierSet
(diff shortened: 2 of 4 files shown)


tetsuo-cpp:
Hmm, the tests seem to be failing to install `requests` in temporary virtual env in 3.10. I'll look into this tomorrow. I think the approach is still fine though.

pip_audit/_dependency_source/requirement.py
```diff
@@ -164,16 +138,38 @@ def _fix_file(self, filename: Path, fix_version: ResolvedFixVersion) -> None:
...
+            try:
+                if not fixed and fix_version.dep in self._collect_cached_deps(
+                    filename, list(reqs.values())
+                ):
+                    logger.warning(
+                        f"added fixed subdependency explicitly to requirements file {filename}: "
+                        f"{fix_version.dep.canonical_name}"
+                    )
+                    f.write("# added by pip-audit to fix subdependency" + os.linesep)
+                    f.write(f"{fix_version.dep.canonical_name}=={fix_version.version}" + os.linesep)
```
woodruffw, line 169: If I'm reading this right, this will produce outputs like this, right?

```
# added by pip-audit to fix subdependency
some-subdep==1.2.3
```

IMO we should do something closer to what `pip-compile` does, for example:

```
some-subdep==1.2.3
    # pip-audit: subdependency fixed via some-dep (4.5.6)
```

Note the indentation of the comment, and the context we provide. OTOH, I'm not sure how much of a pain it'll be to retrieve that context.

We should do the same when a fixed subdep is beneath multiple deps, e.g.:

```
some-subdep==1.2.3
    # pip-audit: subdependency fixed via some-dep (4.5.6), other-dep (7.8.9)
```

pip_audit/_dependency_source/requirement.py
```diff
@@ -164,16 +138,38 @@ def _fix_file(self, filename: Path, fix_version: ResolvedFixVersion) -> None:
...
+            # another.
+            try:
+                if not fixed and fix_version.dep in self._collect_cached_deps(
+                    filename, list(reqs.values())
+                ):
+                    logger.warning(
+                        f"added fixed subdependency explicitly to requirements file {filename}: "
+                        f"{fix_version.dep.canonical_name}"
+                    )
+                    f.write("# added by pip-audit to fix subdependency" + os.linesep)
```
woodruffw, line 168: Tiny nitpick:

```suggestion
                    print("# added by pip-audit to fix subdependency", file=f)
```

here and below, to avoid the manual line separation logic.

pip_audit/_dependency_source/requirement.py
tetsuo-cpp: Good idea. As discussed, I'll do this.

pip_audit/_dependency_source/requirement.py
tetsuo-cpp: [69c42b2](https://github.com/trailofbits/pip-audit/pull/297/commits/69c42b2ee12297eece521e991c866d26362146c1)

CHANGELOG.md
```diff
@@ -24,6 +24,11 @@ All versions prior to 0.0.9 are untracked.
   `pyproject.toml` dependency source
   ([#300](https://github.com/trailofbits/pip-audit/pull/300))
 
+* Vulnerability fixing: the `--fix` flag now works for vulnerabilities found in
```
woodruffw, line 27: This needs to be relocated under the "unreleased" section (this is my fault, due to the `2.3.3` release this morning).

pip_audit/_dependency_source/requirement.py
```diff
@@ -164,15 +138,46 @@ def _fix_file(self, filename: Path, fix_version: ResolvedFixVersion) -> None:
...
+                    if origin_reqs:
+                        logger.warning(
+                            "added fixed subdependency explicitly to requirements file "
+                            f"{filename}: {fix_version.dep.canonical_name}"
+                        )
+                        origin_reqs_formatted = ",".join(
+                            [str(req) for req in sorted(list(origin_reqs), key=lambda x: x.name)]
+                        )
+                        print(
+                            f"# pip-audit: subdependency fixed via {origin_reqs_formatted}",
```
woodruffw, line 175: Indent, to mimic what `pip-compile` does:

```suggestion
                            f"    # pip-audit: subdependency fixed via {origin_reqs_formatted}",
```

woodruffw (approved):

PR #298, opened 2022-06-14
_cli: Add warning when combining --require-hashes and -s osv

### LINKED ISSUE #234: Verify hashes with PyPI when using OSV as a vulnerability service
At the moment, we only verify hashes with using PyPI as the vulnerability service. If we use OSV, we should still query PyPI for the purpose of verifying hashes.

tetsuo-cpp:
Closes #234

### DIFF
--- a/pip_audit/_cli.py
+++ b/pip_audit/_cli.py
@@ -348,6 +348,12 @@ def audit() -> None:
     if args.require_hashes and args.no_deps:
         logger.warning("The --no-deps flag is redundant when used with --require-hashes")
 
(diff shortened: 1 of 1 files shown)


tetsuo-cpp:
```
$ pip-audit --require-hashes -s osv -r requirements0.txt
WARNING:pip_audit._cli:The --require-hashes flag with --service osv only enforces hash presence NOT hash validity. Use --service pypi to enforce hash validity.
```

woodruffw (approved):

PR #300, opened 2022-06-14
Respect `--cache-dir` and other flags when auditing project directories

### LINKED ISSUE #299: `pip-audit --cache-dir=...` is not respected when auditing a project directory
Observed while attempting to run the pip-audit GitHub Action on a directory containing a `pyproject.toml`: we instantiate a `ResolveLibResolver`, but we forget to pass in the user-specified cache dir. As a result it tries to call `python -m pip cache dir`, which can fail on some hosts.

woodruffw:
Fixes #299.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -34,10 +34,6 @@ env/pyvenv.cfg: pyproject.toml
 	./env/bin/python -m pip install --upgrade pip
 	./env/bin/python -m pip install -e .[dev]
 
-.PHONY: run
-run: env/pyvenv.cfg
-	@. env/bin/activate && pip-audit $(ARGS)
-
 .PHONY: lint
 lint: env/pyvenv.cfg
 	. env/bin/activate && \
--- a/Makefile
+++ b/Makefile
@@ -30,7 +30,7 @@ dev: env/pyvenv.cfg
 
 env/pyvenv.cfg: pyproject.toml
 	# Create our Python 3 virtual environment
-	[[ -d env ]] || python3 -m venv env
+	python3 -m venv env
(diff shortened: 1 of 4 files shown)


woodruffw:
Confirmed locally that `pip-audit --cache-dir=/tmp/test .` now works as expected.

Makefile
```diff
@@ -21,23 +21,18 @@ else
...
-
-
-.PHONY: dev
-dev: env/pyvenv.cfg
-
 .PHONY: all
 all:
 	@echo "Run my targets individually!"
 
-.PHONY: run
```
tetsuo-cpp, line 38: What's happening here? I see `make run` is getting removed.

tetsuo-cpp (approved):
LGTM!

woodruffw:
Yep. I realized that the only place it was being used was in the CI and that it wasn’t much better of a UX than directly running the command from within the virtual environment. But I can revert that, if you were using it locally!

tetsuo-cpp:
> Yep. I realized that the only place it was being used was in the CI and that it wasn’t much better of a UX than directly running the command from within the virtual environment. But I can revert that, if you were using it locally! Sent from mobile. Please excuse my brevity.
> […](#)
> On Jun 14, 2022, at 6:16 PM, Alex Cameron ***@***.***> wrote: ﻿ @tetsuo-cpp approved this pull request. LGTM! In Makefile: > .PHONY: all all: @echo "Run my targets individually!" -.PHONY: run What's happening here? I see make run is getting removed. — Reply to this email directly, view it on GitHub, or unsubscribe. You are receiving this because you were assigned.

All good! I wasn't using it, just curious.

PR #309, opened 2022-06-17
workflows: Specify `python-version` for docs and release workflows

tetsuo-cpp:


.github/workflows/ci.yml
```diff
@@ -31,7 +31,7 @@ jobs:
       - uses: actions/checkout@v3
       - uses: actions/setup-python@v4
         with:
-          python-version: "3.7"
+          python-version: "3.x"
```
tetsuo-cpp, line 34: Also changed this to `3.x`. No need to pin to a concrete version I think.

tetsuo-cpp:
I merged in #307 earlier today. One of the incompatibilities with the newer version of the `setup-python` action is that we need to specify `python-version`. I forgot to add that change to the doc and release workflows.

.github/workflows/ci.yml
woodruffw: For the CI specifically, we should probably do a matrix of supported versions here: `3.7`, `3.8`, `3.9`, and `3.10`.

.github/workflows/docs.yml
```diff
@@ -15,6 +15,8 @@ jobs:
       - uses: actions/checkout@v3
 
       - uses: actions/setup-python@v4
+        with:
+          python-version: "3.x"
```
woodruffw, line 19: I think this version of the action supports SemVer qualifiers, so maybe this?

```suggestion
          python-version: ">= 3.7"
```

.github/workflows/release.yml
```diff
@@ -16,6 +16,8 @@ jobs:
     - uses: actions/checkout@v3
 
     - uses: actions/setup-python@v4
+      with:
+        python-version: "3.x"
```
woodruffw, line 20: Same as above.

woodruffw (approved):

PR #323, opened 2022-07-19
OSV, PyPI: Do not use "details" field for vuln summaries

### LINKED ISSUE #314: Slightly broken Markdown formatting when used with `--desc`
This is a relatively minor bug: the combination of `--desc` and `--format=markdown` might result in a broken Markdown table, since `--desc` can include multiple lines of embedded Markdown from the vulnerability service.

For example, from the self-test on our GitHub Action: https://github.com/trailofbits/gh-action-pip-audit/actions/runs/2592038764

Screen Shot 2022-06-30 at 2 52 43 PM

---

I see two possible solutions (but there are probably others!):

* Forbid the combination of `--desc` and `--format=markdown`
    * Pros: Arguably the most correct, since `--desc` is explicitly allowed to contain Markdown per the OSV spec on vulnerability descriptions
    * Cons: Makes the Markdown format a little less useful.
* "Collapse" the Markdown into a single line
    * Pros: Preserves descriptions in the Markdown table, which aren't multiple lines 99% of the time. 
    * Cons: Will slightly mis-render any Markdown descriptions that contain multiple lines, *if* those multiple lines are semantically important (e.g. headings)

I'm slightly leaning towards the second.

woodruffw:
The "details" field is frequently multiline and OSV allows it to be Markdown formatted, which causes problems for the surrounding columnar and Markdown formats. The correct field in our context is the "summary" field, which is specified as a short plaintext string.

Closes #314.

### DIFF
--- a/pip_audit/_service/osv.py
+++ b/pip_audit/_service/osv.py
@@ -90,9 +90,19 @@ class OsvService(VulnerabilityService):
 
             id = vuln["id"]
 
+            # The summary is intended to be shorter, so we prefer it over
+            # details, if present. However, neither is required.
             description = vuln.get("summary")
             if description is None:
+                description = vuln.get("details")
+            if description is None:
                 description = "N/A"
+
+            # The "summary" field should be a single line, but "details" might
+            # be multiple (Markdown-formatted) lines. So, we normalize our
+            # description into a single line (and potentially break the Markdown
+            # formatting in the process).
+            description = description.replace("\n", " ")
 
             aliases = set(vuln.get("aliases", []))
 
--- a/pip_audit/_service/osv.py
+++ b/pip_audit/_service/osv.py
@@ -90,11 +90,7 @@ class OsvService(VulnerabilityService):
 
             id = vuln["id"]
 
-            # The summary is intended to be shorter, so we prefer it over
-            # details, if present. However, neither is required.
             description = vuln.get("summary")
-            if description is None:
-                description = vuln.get("details")
             if description is None:
                 description = "N/A"
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,12 @@ All versions prior to 0.0.9 are untracked.
 
 ## [Unreleased]
 
+### Fixed
+
+* CLI: the `--format=markdown` and `--format=columns` output formats are no
+  longer broken by long vulnerability descriptions from the OSV and PyPI
+  vulnerability sources ([#323](https://github.com/trailofbits/pip-audit/pull/323))
+
 ## [2.4.1]
 
 ### Fixed
--- a/pip_audit/_service/pypi.py
+++ b/pip_audit/_service/pypi.py
@@ -117,8 +117,12 @@ class PyPIService(VulnerabilityService):
             # The ranges aren't guaranteed to come in chronological order
             fix_versions.sort()
 
+            description = v.get("summary")
+            if description is None:
+                description = "N/A"
+
             results.append(
-                VulnerabilityResult(v["id"], v["details"], fix_versions, set(v["aliases"]))
+                VulnerabilityResult(v["id"], description, fix_versions, set(v["aliases"]))
             )
 
         return spec, results
--- a/pip_audit/_service/pypi.py
+++ b/pip_audit/_service/pypi.py
@@ -119,7 +119,16 @@ class PyPIService(VulnerabilityService):
 
             description = v.get("summary")
             if description is None:
+                description = v.get("details")
+
+            if description is None:
                 description = "N/A"
+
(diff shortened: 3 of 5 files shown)


pip_audit/_service/osv.py
```diff
@@ -90,11 +90,7 @@ def query(self, spec: Dependency) -> Tuple[Dependency, List[VulnerabilityResult]
 
             id = vuln["id"]
 
-            # The summary is intended to be shorter, so we prefer it over
-            # details, if present. However, neither is required.
             description = vuln.get("summary")
-            if description is None:
-                description = vuln.get("details")
```
di, line 97: I kind of wonder if we should fall back on a portion of the description, maybe everything up until the first newline or something?

woodruffw: That'll work in many cases, although there'll be a handful where it'll produce something relatively useless (like a Markdown `## header`). Overall probably worth it, so I'll add it.

woodruffw: Actually, maybe we should just fall back on `details`, but with newlines replaced with spaces? That'll result in a big blob of text in the worst case, but it'll avoid the situation where we render the first line and it doesn't happen to be useful.

di: I think that's a good idea.

woodruffw: That's much nicer 🙂 


Name | Version | ID | Fix Versions | Description
--- | --- | --- | --- | ---
twisted | 18.9.0 | PYSEC-2019-128 | 19.2.1 | In Twisted before 19.2.1, twisted.web did not validate or sanitize URIs or HTTP methods, allowing an attacker to inject invalid characters such as CRLF.
twisted | 18.9.0 | PYSEC-2020-260 | 20.3.0rc1 | In Twisted Web through 19.10.0, there was an HTTP request splitting vulnerability. When presented with a content-length and a chunked encoding header, the content-length took precedence and the remainder of the request body was interpreted as a pipelined request.
twisted | 18.9.0 | PYSEC-2019-129 | 19.7.0rc1 | In words.protocols.jabber.xmlstream in Twisted through 19.2.1, XMPP support did not verify certificates when used with TLS, allowing an attacker to MITM connections.
twisted | 18.9.0 | PYSEC-2020-259 | 20.3.0rc1 | In Twisted Web through 19.10.0, there was an HTTP request splitting vulnerability. When presented with two content-length headers, it ignored the first header. When the second content-length value was set to zero, the request body was interpreted as a pipelined request.
twisted | 18.9.0 | PYSEC-2022-27 | 22.1.0 | twisted is an event-driven networking engine written in Python. In affected versions twisted exposes cookies and authorization headers when following cross-origin redirects. This issue is present in the `twited.web.RedirectAgent` and `twisted.web. BrowserLikeRedirectAgent` functions. Users are advised to upgrade. There are no known workarounds.
twisted | 18.9.0 | PYSEC-2022-195 | 22.4.0 | Twisted is an event-based framework for internet applications, supporting Python 3.6+. Prior to version 22.4.0rc1, the Twisted Web HTTP 1.1 server, located in the `twisted.web.http` module, parsed several HTTP request constructs more leniently than permitted by RFC 7230. This non-conformant parsing can lead to desync if requests pass through multiple HTTP parsers, potentially resulting in HTTP request smuggling. Users who may be affected use Twisted Web's HTTP 1.1 server and/or proxy and also pass requests through a different HTTP server and/or proxy. The Twisted Web client is not affected. The HTTP 2.0 server uses a different parser, so it is not affected. The issue has been addressed in Twisted 22.4.0rc1. Two workarounds are available: Ensure any vulnerabilities in upstream proxies have been addressed, such as by upgrading them; or filter malformed requests by other means, such as configuration of an upstream proxy.
twisted | 18.9.0 | GHSA-32gv-6cf3-wcmq | 19.10.0 | ### Impact Twisted web servers that utilize the optional HTTP/2 support suffer from the following flow-control related vulnerabilities:  Ping flood: https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2019-9512 Reset flood: https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2019-9514 Settings flood: https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2019-9515  A Twisted web server supports HTTP/2 requests if you've installed the [`http2` optional dependency set](https://twistedmatrix.com/documents/19.2.0/installation/howto/optional.html).  ### Workarounds There are no workarounds.  ### References https://github.com/Netflix/security-bulletins/blob/master/advisories/third-party/2019-002.md  ### For more information If you have any questions or comments about this advisory: * Open an issue in [Twisted's Trac](https://twistedmatrix.com/trac/)

Issue #171, opened 2021-12-03
Hide spinner and progress when done


layday:
I ran `pip-audit` and tabbed away from the terminal. When I tabbed back it said:

```
$ pip-audit
| Auditing yarl (1.7.2)
No known vulnerabilities found
```

This confused me, thinking it had only audited yarl, so I re-ran it.  This is a fairly minor UX nit, but if there's an easy fix, I think it'd be better to hide the progress text once auditing is finished.

di:
I agree, this was bothering me when I was trying to show examples of `pip-audit` output.

woodruffw:
Thanks for the report! Yeah, we can definitely improve this.

Hiding the spinner/progress entirely might be a little tricky to do with just our current use of the `progress` package, but we could improve the final text to make it clear that we've performed the entire audit. Something like this, perhaps:

```
$ pip-audit
| Auditing yarl (1.7.2)
123 packages audited; no known vulnerabilities found
```

Thoughts?

di:
@woodruffw Can't we just do one last spinner "update" with the results?

woodruffw:
> Can't we just do one last spinner "update" with the results?

Yeah, I think that would work. Are you thinking that we'd remove the current trailing line in that case, or keep it as a duplicate?

di:
Remove it I guess, but that would probably be dependent on the state of `--progress-spinner`.

woodruffw:
Yeah, that's the corner case I was thinking of. I'll play around with it and make some demos with the PR.

layday:
I don't have a lot of experience with manipulating the terminal output and I've not spent very long looking at how `progress` works, but in `AuditSpinnerfinalise`, instead of calling the spinner's `finish` method I think that you should be able to:

```py
self.writeln('')  # Replace printed characters with whitespace
self.file.write('\r')  # Reset the cursor
self.file.flush()
```

woodruffw:
Oh, interesting. I had pessimistically assumed that it would be worse than that 😅. I'll test it out.

layday:
Thanks for the quick fix!

woodruffw:
You're welcome; thanks for the report! Let us know if you have any other issues.

PR #329, opened 2022-07-22
Makefile: Tweak linting target and add new `reformat` target

tetsuo-cpp:
The ideas from #171 are good so I figure we should port them across to `pip-audit`.


woodruffw (approved):
LGTM! Good call.

PR #343, opened 2022-07-29
pyproject: Use recursive extras to optimize the build process

woodruffw:
This breaks the `dev` extra into two sub-extras, `test` and `lint`, allowing us to install only the subset of development dependencies needed for each. This in turn allows us to optimize our CI by installing only the dependencies required for each CI job.

Also: I'm using this PR to test `gitsign`, which is why the commits are marked as "Unverified". But they're signed via Sigstore + my GitHub identity!

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -24,7 +24,7 @@ jobs:
         with:
           python-version: ${{ matrix.python }}
       - name: test
-        run: make test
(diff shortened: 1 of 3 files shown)


woodruffw:
Improvements: 24s for the `lint` job vs. 33s on `main`; 1m 53s for the `test (3.7)` job vs. 2m 2s on `main`.

tetsuo-cpp (approved):
LGTM!

Issue #340, opened 2022-07-28
CalledProcessError when specifiying a package in requirements


ccsalway:
Dont understand whats happening here.. I can run pip-audit against an installed pacakge, but naming the same package in the requriements.txt file renders and error...

```
% python3 -m venv --clear .venv && . .venv/bin/activate
% python -m pip install -Uq pip wheel setuptools pip-audit
% pip freeze
CacheControl==0.12.11
certifi==2022.6.15
charset-normalizer==2.1.0
commonmark==0.9.1
cyclonedx-python-lib==2.7.0
html5lib==1.1
idna==3.3
lockfile==0.12.2
msgpack==1.0.4
packageurl-python==0.10.0
packaging==21.3
pip-api==0.0.30
pip-requirements-parser==31.2.0
pip_audit==2.4.3
Pygments==2.12.0
pyparsing==3.0.9
requests==2.28.1
resolvelib==0.8.1
rich==12.5.1
six==1.16.0
sortedcontainers==2.4.0
toml==0.10.2
typing_extensions==4.3.0
urllib3==1.26.11
webencodings==0.5.1

% python -m pip_audit        
No known vulnerabilities found

% pip install -Uq flask==0.5
% python -m pip_audit   
Found 2 known vulnerabilities in 1 package
Name  Version ID             Fix Versions
----- ------- -------------- ------------
flask 0.5     PYSEC-2019-179 1.0
flask 0.5     PYSEC-2018-66  0.12.3

% echo "flask==0.5" > requirements.txt
% python -m pip_audit -r requirements.txt 
WARNING:cachecontrol.controller:Cache entry deserialization failed, entry ignored
Traceback (most recent call last):
  File "/Library/Developer/CommandLineTools/Library/Frameworks/Python3.framework/Versions/3.8/lib/python3.8/runpy.py", line 194, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/Library/Developer/CommandLineTools/Library/Frameworks/Python3.framework/Versions/3.8/lib/python3.8/runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "/private/var/folders/7b/w1fc1k292kgdj7g01ry8vf_40000gn/T/tmp.TBMsMgV1/.venv/lib/python3.8/site-packages/pip_audit/__main__.py", line 8, in <module>
    audit()
[...]

PR #344, opened 2022-07-29
resolvelib, test: handle venv creation errors gracefully

woodruffw:
This is a _partial_ solution for #340 -- it converts an uncontrolled exception into one we control, which a later PR can then use to add some appropriate end-user messaging (such as a custom message telling the user that their system Python is badly borked).

Some follow-up options (non-exhaustive) are listed in https://github.com/trailofbits/pip-audit/issues/340#issuecomment-1199800600.

See #340.


woodruffw:
@tetsuo-cpp mind taking a look at this when you have a moment?

tetsuo-cpp (approved):
LGTM! Sorry for the delay.

PR #357, opened 2022-08-24
resolvelib: Consider a project "not found" if its PyPI project page has no links

### LINKED ISSUE #356: Dependency resolution errors for URL requirements that don't exist in PyPI
I noticed this as part of debugging #354. Context in https://github.com/trailofbits/pip-audit/issues/354#issuecomment-1225300272.

tetsuo-cpp:
Closes #356

### DIFF
--- a/pip_audit/_dependency_source/resolvelib/pypi_provider.py
+++ b/pip_audit/_dependency_source/resolvelib/pypi_provider.py
@@ -213,7 +213,10 @@ def get_project_from_index(
     response.raise_for_status()
     data = response.content
     doc = html5lib.parse(data, namespaceHTMLElements=False)
-    for i in doc.findall(".//a"):
+    links = doc.findall(".//a")
+    if not links:
+        raise PyPINotFoundError
+    for i in links:
         url = i.attrib["href"]
         py_req = i.attrib.get("data-requires-python")
         # Skip items that need a different Python version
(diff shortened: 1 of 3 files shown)


pip_audit/_dependency_source/resolvelib/pypi_provider.py
```diff
@@ -213,7 +213,10 @@ def get_project_from_index(
     response.raise_for_status()
     data = response.content
     doc = html5lib.parse(data, namespaceHTMLElements=False)
-    for i in doc.findall(".//a"):
+    links = doc.findall(".//a")
+    if not links:
+        raise PyPINotFoundError
```
woodruffw, line 218: Tiny nit, but it might make sense to specialize the error slightly here (since PyPI *does* have a project registered for the name, but the project is just missing releases).

woodruffw (approved):
One nonblocking nit, otherwise LGTM!

pip_audit/_dependency_source/resolvelib/pypi_provider.py
woodruffw: Actually, thinking about this more, I don't think it makes sense to distinguish these error states. They're effectively same from the `pip` end user's perspective, so they should be the same from a `pip-audit` user's perspective.

PR #359, opened 2022-08-24
requirement: Remove incorrect assertion around URL requirements without egg fragment

### LINKED ISSUE #350: Bug: assertion tripped when auditing a URL-style requirement without egg marker
I noticed this while root-causing #349.

The following:

```bash
pip-audit -r <(echo 'git+https://github.com/unbit/uwsgi.git@1bb9ad77c6d2d310c2d6d1d9ad62de61f725b824')
```

fails with:

```
  File "/Users/william/devel/pip-audit/env/bin/pip-audit", line 8, in <module>
    sys.exit(audit())
  File "/Users/william/devel/pip-audit/pip_audit/_cli.py", line 432, in audit
    for (spec, vulns) in auditor.audit(source):
  File "/Users/william/devel/pip-audit/pip_audit/_audit.py", line 66, in audit
    for dep, vulns in self._service.query_all(specs):
  File "/Users/william/devel/pip-audit/pip_audit/_service/interface.py", line 150, in query_all
    for spec in specs:
  File "/Users/william/devel/pip-audit/pip_audit/_dependency_source/requirement.py", line 104, in collect
    assert isinstance(req, EditableRequirement)
AssertionError
```

...which is:

```python
for req in rf.requirements:
    if req.req is None:
        # For editable requirements that don't have an egg fragment that lists the
        # the package name and version, `pip-requirements-parser` won't attach a
        # `Requirement` object to the `InstallRequirement`.
        #
        # In this case, we can't audit the dependency so we should signal to the
        # caller that we're skipping it.
        assert isinstance(req, EditableRequirement)
        yield SkippedDependency(
            name=req.requirement_line.line,
            skip_reason="could not deduce package/specifier pair from requirement, "
            "please specify them with #egg=your_package_name==your_package_version",
        )
        continue
```

The `assert` here feels spurious -- we should support URL requirements that aren't marked as editable.

CC @tetsuo-cpp for thoughts, since I think this was your change.

tetsuo-cpp:
Closes #350

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,12 @@ All versions prior to 0.0.9 are untracked.
 
 ## [Unreleased]
 
+### Fixed
+
+* Fixed an incorrect assertion triggering for non-editable URL requirements that
+  don't have an egg fragment
+  ([#359](https://github.com/trailofbits/pip-audit/pull/359))
+
 ## [2.4.3]
 
 ### Fixed
--- a/pip_audit/_dependency_source/requirement.py
+++ b/pip_audit/_dependency_source/requirement.py
@@ -14,12 +14,7 @@ from typing import IO, Dict, Iterator, List, Set, Tuple, cast
 from packaging.requirements import Requirement
 from packaging.specifiers import SpecifierSet
 from packaging.version import Version
-from pip_requirements_parser import (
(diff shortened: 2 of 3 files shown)


woodruffw (approved):
LGTM!

PR #360, opened 2022-08-25
README: sec model: clarify what we detect

### LINKED ISSUE #358: How should the sdist/wheel distinction be handled and communicated?
First, pip-audit is a really good idea, and I'm going to look at how best to start using it in projects I work on. The python community needs this kind of tooling, so thank you for developing and maintaining this. However, I can see a major problem with pip-audit, in that it seems to assume that a vulnerability is tied to a specific package and version, rather than a specific artefact. This has issues where wheels bundle other, non-Python libraries, and where vulnerabilities in those libraries affect only the pre-compiled wheels, or even only a subset of the pre-compiled wheels. Assuming python package `A` bundles vulnerable version of non-Python library `B` in its wheel, then given an environment with `A`, pip-audit currently appears to have two options:
 * Tell the user package `A` is vulnerable, even if vulnerable version of `B` is not used (e.g. a fixed version of `B` is installed by another tool such as conda, or `B` can be replaced by `C` which is API compatible with `B` but does not share `B`'s vulnerability). This would result in a "pip-audit cries woof" situation like npm audit, which means people ignore pip-audit. It would also likely increase the pip vs conda division.
 * Do not tell the user package `A` is vulnerable, people find out pip-audit did not tell them that they installed a vulnerable artefact, and complain "pip" does not work (and ecosystem Y is better).

Neither of these seem ideal, and I can't see a way around them as long as sdists and wheels have the same version number (having different version numbers is a way to resolve a related issue around distribution-packaged versions where a fix has been applied), or somehow the vulnerabilities are tied to artefacts. I suspect the best current solution would be to have clear documentation around what pip-audit does in such cases, and what users can do to avoid hitting this issue (e.g. what other audit tools can be used to complement pip-audit).

woodruffw:
Closes #358.

cc @aragilar for feedback!

### DIFF
--- a/README.md
+++ b/README.md
@@ -412,6 +412,14 @@ and purposes, `pip-audit -r INPUT` is functionally equivalent to
 `pip install -r INPUT`, with a small amount of **non-security isolation** to
 avoid conflicts with any of your local environments.
 
+`pip-audit` is first and foremost a auditing tool for *Python* packages.
+You **must not** assume that `pip-audit` will detect or flag "transitive"
+vulnerabilities that might be exposed through Python packages, but are not
+actually part of the package itself. For example, `pip-audit`'s vulnerability
+information sources are unlikely to include an advisory for a vulnerable shared
+library that a popular Python package *might* use, since the Python package's
+version is not strongly connected to the shared library's version.
+
 ## Licensing
 
 `pip-audit` is licensed under the Apache 2.0 License.


tetsuo-cpp (approved):
Nice, I like how this is worded. LGTM!

Thanks @aragilar for pointing this out.

PR #377, opened 2022-10-03
workflows/release: update gh-action-sigstore-python ref

woodruffw:
Now that we've reorg'd under the Sigstore org.


tetsuo-cpp (approved):
LGTM!

PR #386, opened 2022-10-22
osv: skip results marked as "withdrawn"

### LINKED ISSUE #385: Skip vulnerability reports that are marked as "withdrawn"
The OSV format supports the `withdrawn` field, which indicates that a report has been withdrawn and is no longer valid. 

`pip-audit` should respect that field and ignore any reports that contain it, so long as the field's timestamp is not after the current time.

woodruffw:
~~WIP; needs tests.~~

Closes #385. 

### DIFF
--- a/pip_audit/_service/osv.py
+++ b/pip_audit/_service/osv.py
@@ -90,6 +90,12 @@ class OsvService(VulnerabilityService):
 
             id = vuln["id"]
 
+            # If the vulnerability has been withdrawn, we skip it entirely.
+            withdrawn_at = vuln.get("withdrawn")
+            if withdrawn_at is not None:
+                logger.debug(f"vuln {id} marked as withdrawn at {withdrawn_at}; skipping")
+                continue
+
             # The summary is intended to be shorter, so we prefer it over
             # details, if present. However, neither is required.
             description = vuln.get("summary")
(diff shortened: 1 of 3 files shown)


pip_audit/_service/osv.py
```diff
@@ -90,6 +90,12 @@ def query(self, spec: Dependency) -> Tuple[Dependency, List[VulnerabilityResult]
 
             id = vuln["id"]
 
+            # If the vulnerability has been withdrawn, we skip it entirely.
+            withdrawn_at = vuln.get("withdrawn")
+            if withdrawn_at is not None:
+                logger.debug(f"OSV vuln entry '{id}' marked as withdrawn at {withdrawn_at}")
```
jenstroeger, line 96: Without being asked for my opinion, but shouldn’t logging use lazy formatting for performance reasons, instead of f-strings?
```python
logger.debug(f"OSV vuln entry '%s' marked as withdrawn at %s", id, withdrawn_at)
```
(See also [Optimization](https://docs.python.org/3/howto/logging.html#optimization) and [issue 46200](https://bugs.python.org/issue46200), which might not apply in this particular case.)

woodruffw: Normally I would agree, but I think this is a premature optimization in this context (and in `pip-audit` as a whole): we're fundamentally network and resolver bound, so a couple of f-strings probably aren't doing anything too bad.

That being said, we can reevaluate that position if tracing shows a significant hit.

di (approved):

PR #402, opened 2022-11-04
Add support for Poetry

### LINKED ISSUE #84: Support Poetry via `poetry.lock`
Breakout from #73.

orsinium:
If the project path is specified in CLI, read dependencies from `poetry.lock` file (if available). 

I'm quite happy with `collect`. Since this is a lock file, we can safely skip dependency resolution, as we do for `pip` source. And the format of the file is quite simple, so we don't depend on poetry for reading it. Cool!

Autofix is not supported yet because poetry needs hashes to be updated, see review comments.

I checked the change manually on a few quite big poetry projects, works like a charm.

Close #84

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,7 +37,7 @@ with support from Google. This is not an official Google or Trail of Bits produc
 
 ## Features
 
-* Support for auditing local environments and requirements-style files
+* Support for auditing local environments, requirements-style files, and poetry.lock files.
 * Support for multiple vulnerability services
   ([PyPI](https://warehouse.pypa.io/api-reference/json.html#known-vulnerabilities),
   [OSV](https://osv.dev/docs/))
--- a/README.md
+++ b/README.md
@@ -37,7 +37,7 @@ with support from Google. This is not an official Google or Trail of Bits produc
 
 ## Features
 
-* Support for auditing local environments, requirements-style files, and poetry.lock files.
+* Support for auditing local environments, requirements-style files, and poetry.lock files
 * Support for multiple vulnerability services
   ([PyPI](https://warehouse.pypa.io/api-reference/json.html#known-vulnerabilities),
   [OSV](https://osv.dev/docs/))
--- a/pip_audit/_cli.py
+++ b/pip_audit/_cli.py
@@ -17,6 +17,7 @@ from pip_audit._dependency_source import (
     PYPI_URL,
     DependencySource,
     PipSource,
+    PoetrySource,
     PyProjectSource,
     RequirementSource,
     ResolveLibResolver,
@@ -330,8 +331,14 @@ def _dep_source_from_project_path(
     project_path: Path, resolver: ResolveLibResolver, state: AuditState
 ) -> DependencySource:  # pragma: no cover
     # Check for a `pyproject.toml`
+    poetry_lock = project_path / "poetry.lock"
+    if poetry_lock.is_file():
+        logger.debug("using PoetrySource as dependency source")
+        return PoetrySource(path=poetry_lock)
+
     pyproject_path = project_path / "pyproject.toml"
     if pyproject_path.is_file():
+        logger.debug("using PyProjectSource as dependency source")
         return PyProjectSource(pyproject_path, resolver, state)
 
     # TODO: Checks for setup.py and other project files will go here.
--- a/pip_audit/_cli.py
+++ b/pip_audit/_cli.py
@@ -330,7 +330,6 @@ def _parse_args(parser: argparse.ArgumentParser) -> argparse.Namespace:  # pragm
 def _dep_source_from_project_path(
     project_path: Path, resolver: ResolveLibResolver, state: AuditState
 ) -> DependencySource:  # pragma: no cover
-    # Check for a `pyproject.toml`
     poetry_lock = project_path / "poetry.lock"
     if poetry_lock.is_file():
         logger.debug("using PoetrySource as dependency source")
--- a/pip_audit/_dependency_source/poetry.py
+++ b/pip_audit/_dependency_source/poetry.py
@@ -11,7 +11,7 @@ from typing import Iterator
 import toml
 from packaging.version import InvalidVersion, Version
 
-from pip_audit._dependency_source import DependencySource
+from pip_audit._dependency_source import DependencyFixError, DependencySource
 from pip_audit._fix import ResolvedFixVersion
 from pip_audit._service import Dependency, ResolvedDependency, SkippedDependency
 
@@ -41,7 +41,6 @@ class PoetrySource(DependencySource):
                     "Package has invalid version and could not be audited: "
                     f"{name} ({package['version']})"
                 )
-                logger.debug(skip_reason)
                 yield SkippedDependency(name=name, skip_reason=skip_reason)
             else:
                 yield ResolvedDependency(name=name, version=version)
@@ -55,4 +54,4 @@ class PoetrySource(DependencySource):
         Note that poetry ignores the version we want to update to,
         and goes straight to the latest version allowed in metadata.
         """
-        raise NotImplementedError("fix is not supported for poetry yet")  # pragma: no cover
+        raise DependencyFixError("fix is not supported for poetry yet")  # pragma: no cover
--- a/pip_audit/_dependency_source/poetry.py
+++ b/pip_audit/_dependency_source/poetry.py
@@ -36,7 +36,7 @@ class PoetrySource(DependencySource):
             name = package["name"]
             try:
                 version = Version(package["version"])
-            except InvalidVersion:  # pragma: no cover
+            except InvalidVersion:
                 skip_reason = (
                     "Package has invalid version and could not be audited: "
                     f"{name} ({package['version']})"
@@ -55,4 +55,4 @@ class PoetrySource(DependencySource):
         Note that poetry ignores the version we want to update to,
         and goes straight to the latest version allowed in metadata.
         """
-        raise NotImplementedError("fix is not supported for poetry yet")
+        raise NotImplementedError("fix is not supported for poetry yet")  # pragma: no cover
--- a/pip_audit/_dependency_source/poetry.py
+++ b/pip_audit/_dependency_source/poetry.py
@@ -4,8 +4,6 @@ Collect dependencies from `poetry.lock` files.
 from __future__ import annotations
 
 import logging
-import subprocess
-import sys
 from dataclasses import dataclass
 from pathlib import Path
 from typing import Iterator
@@ -57,8 +55,4 @@ class PoetrySource(DependencySource):
         Note that poetry ignores the version we want to update to,
         and goes straight to the latest version allowed in metadata.
         """
-        subprocess.run(
-            [sys.executable, "-m", "poetry", "update", "--lock", fix_version.dep.name],
-            cwd=self.path.parent,
-            stdout=subprocess.DEVNULL,
-        ).check_returncode()
+        raise NotImplementedError("fix is not supported for poetry yet")
--- a/pip_audit/_dependency_source/poetry.py
+++ b/pip_audit/_dependency_source/poetry.py
@@ -38,7 +38,7 @@ class PoetrySource(DependencySource):
             name = package["name"]
             try:
                 version = Version(package["version"])
-            except InvalidVersion:
+            except InvalidVersion:  # pragma: no cover
                 skip_reason = (
                     "Package has invalid version and could not be audited: "
                     f"{name} ({package['version']})"
--- a/pip_audit/_dependency_source/poetry.py
+++ b/pip_audit/_dependency_source/poetry.py
@@ -1,0 +1,64 @@
+"""
+Collect dependencies from `poetry.lock` files.
+"""
+from __future__ import annotations
+
+import logging
+import subprocess
+import sys
+from dataclasses import dataclass
+from pathlib import Path
+from typing import Iterator
+
+import toml
+from packaging.version import InvalidVersion, Version
+
+from pip_audit._dependency_source import DependencySource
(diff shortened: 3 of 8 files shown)


woodruffw:
Thanks a ton for the PR @orsinium! I'll do a review in a moment.

pip_audit/_dependency_source/poetry.py
```diff
@@ -0,0 +1,64 @@
...
+        """
+        Collect all of the dependencies discovered by this `PoetrySource`.
+        """
+        with self.path.open("r") as stream:
+            packages = toml.load(stream)
+        for package in packages["package"]:
+            name = package["name"]
+            try:
+                version = Version(package["version"])
+            except InvalidVersion:  # pragma: no cover
```
woodruffw, line 41: Let's get coverage for this branch! I believe the other dependency sources have similar unit tests for invalid versions.

pip_audit/_dependency_source/poetry.py
```diff
@@ -0,0 +1,64 @@
...
+                skip_reason = (
+                    "Package has invalid version and could not be audited: "
+                    f"{name} ({package['version']})"
+                )
+                logger.debug(skip_reason)
```
woodruffw, line 46: IMO no need for a `debug` here -- we'll immediately surface this reason as a skipped dependency anyways.

pip_audit/_dependency_source/poetry.py
```diff
@@ -0,0 +1,64 @@
...
+        Requires poetry to be installed in the same env.
+
+        Note that poetry ignores the version we want to update to,
+        and goes straight to the latest version allowed in metadata.
+        """
+        subprocess.run(
+            [sys.executable, "-m", "poetry", "update", "--lock", fix_version.dep.name],
+            cwd=self.path.parent,
+            stdout=subprocess.DEVNULL,
+        ).check_returncode()
```
woodruffw, line 64: Unfortunately, I don't think this is an acceptable tradeoff: we want users of `pip-audit` to trust that `pip-audit --fix` actually does the fix upgrades that the vulnerability reports suggest, not that it bumps to whatever `poetry update` thinks the latest acceptable version is.

Similarly, implicitly requiring users to have `poetry` installed is probably not an acceptable requirement for `pip-audit` -- we're developing it so that it *can* be integrated into `pip` at a later date, so executing non-`pip` programs is probably a non-starter.

The way I see it, we have two options here:

* We can try and manually modify the lockfile ourselves. You mentioned a bunch of hashes; assuming they're distribution hashes of the kind that `requirements.txt` supports, we could retrieve them from PyPI. I'm not sure how difficult this would be, since I'm not too familiar with Poetry or its internals 😅 
* We could return an error when a user tries to use `--fix` with a `PoetrySource`. This would be unlike any of our other current packages, but IMO it's not the worst choice if there's no straightforward/reliable way to perform the upgrades ourselves.

cc @di for thoughts.

pyproject.toml
```diff
@@ -44,6 +44,7 @@ requires-python = ">=3.7"
 [project.optional-dependencies]
 test = [
     "coverage[toml]",
+    "poetry",
```
woodruffw, line 47: xref the comment about depending on `poetry` -- I don't think we want this dep, even as a test-only dep.

test/dependency_source/test_poetry.py
```diff
@@ -0,0 +1,63 @@
...
+        metadata = dedent(metadata)
+        metadata += "\n".join(deps)
+        (tmp_path / "pyproject.toml").write_text(metadata)
+        cmd = [sys.executable, "-m", "poetry", "lock", "--no-update"]
+        subprocess.run(cmd, cwd=tmp_path).check_returncode()
+        lock_path = tmp_path / "poetry.lock"
+        assert lock_path.exists()
+        return lock_path
+
+    return callback
```
woodruffw, line 39: Could you move this into `conftest.py`? That's where the rest of our test fixtures live 🙂

woodruffw (changes requested):
This is a really great start! Thanks a ton, again, for tackling this feature.

I've highlighted a couple of small nits, plus a big design question that needs to be resolved.

pip_audit/_dependency_source/poetry.py
woodruffw: I looked into this a little more, and it looks like the hashes in question are indeed the distribution hashes from PyPI. So we'd have to do some network interaction for Poetry fixes, but that's not a huge deal.

More generally, the hashed bits look like this:

```toml
[metadata]
lock-version = "1.1"
python-versions = "^3.7"
content-hash = "7db62de4ee4f2831829fc08b6a3d190f177c62d5af7676e20fb228bb58274850"

[metadata.files]
attrs = [
    {file = "attrs-22.1.0-py2.py3-none-any.whl", hash = "sha256:86efa402f67bf2df34f51a335487cf46b1ec130d02b8d39fd248abfd30da551c"},
    {file = "attrs-22.1.0.tar.gz", hash = "sha256:29adc2665447e5191d0e7c568fde78b21f9672d344281d0c6e1ab085429b22b6"},
]
```

So we have to update each hash under each package under `metadata.files`, as well as the `content-hash`...maybe? I don't know what that one represents.

woodruffw: Oh, `content-hash` seems to be a SHA256 of the `pyproject.toml`. We shouldn't be modifying that, so I _think_ that can be left unchanged?

pip_audit/_dependency_source/poetry.py
orsinium: Are you sure? This code is copy-pasted from pip dependency source. Do you want me to remove it from PipSource code as well?

pip_audit/_dependency_source/poetry.py
orsinium: Let's leave a proper implementation out of scope. For now, I'll go with the second option and raise NotImplementedError TBH, I personally don't see a use case for `--fix` at all. If dependencies are easy to keep up-to-date, just keep the up-to-date on a daily basis, with dependabot or similar. If there are hard to keep up-to-date, autoupgrading them isn't an option. Even if some people really do use `--fix`, definitely not a majority of users. So, I think that can wait a bit.

pyproject.toml
orsinium: Alrighty, I'll drop it for now. But I suggest having integration tests for poetry at some point. When I was doing [dephell](https://github.com/dephell/dephell/), I remember poetry support breaking quite often, mostly because of new features that users expected dephell to support too (like a list of constraints for a single dependency). But poetry should be more stable now, I think.

pip_audit/_dependency_source/poetry.py
orsinium: > So we have to update each hash under each package under metadata.files, as well as the content-hash...maybe? I don't know what that one represents.

It's not only updating hashes of files but also updating the list of available distributions for the release. It stretches what you have to put into `ResolvedFixVersion` quite a bit. The PyPI JSON API provides that info, but then you'll need to go and request it. Currently, PoetrySource is quite simple and self-contained. Don't make another dephell ;)

pip_audit/_dependency_source/poetry.py
orsinium: Fixed ✅

pip_audit/_dependency_source/poetry.py
orsinium: Removed the `fix` code, updated the PR description ✅

pyproject.toml
orsinium: Rewrote tests, removed poetry from dependencies ✅

test/dependency_source/test_poetry.py
orsinium: Removed the fixture altogether ✅

pip_audit/_dependency_source/poetry.py
woodruffw: Yeah, I think we can remove both -- I didn't realize we had it in the other either :slightly_smiling_face:

README.md
```diff
@@ -37,7 +37,7 @@ with support from Google. This is not an official Google or Trail of Bits produc
 
 ## Features
 
-* Support for auditing local environments and requirements-style files
+* Support for auditing local environments, requirements-style files, and poetry.lock files.
```
woodruffw, line 40: Nit: remove period for consistency with other bullets

```suggestion
* Support for auditing local environments, requirements-style files, and poetry.lock files
```

pip_audit/_dependency_source/poetry.py
```diff
@@ -0,0 +1,58 @@
...
+    def fix(self, fix_version: ResolvedFixVersion) -> None:
+        """
+        Fixes a dependency version for this `PoetrySource`.
+
+        Requires poetry to be installed in the same env.
+
+        Note that poetry ignores the version we want to update to,
+        and goes straight to the latest version allowed in metadata.
+        """
+        raise NotImplementedError("fix is not supported for poetry yet")  # pragma: no cover
```
woodruffw, line 58: LGTM. I think we should also handle the combination of `--fix` and `PoetrySource` in the CLI itself, so that we don't expose this bare exception to users.

test/dependency_source/test_poetry.py
```diff
@@ -0,0 +1,34 @@
+from __future__ import annotations
+
+from pathlib import Path
+
+from packaging.version import Version
+
+from pip_audit._dependency_source import PoetrySource
+from pip_audit._service import ResolvedDependency, SkippedDependency
+
+TEST_DATA_PATH = Path(__file__).parent / "data"
```
woodruffw, line 10: Another tiny nitpick: could you turn this into an `asset` or `test_data` fixture and put it in `conftest.py`?

Here's the pattern we use on a handful of other projects:

```python
# conftest.py

_ASSETS = (Path(__file__).parent / "assets").resolve()
assert _ASSETS.is_dir()

@pytest.fixture
def asset():
    def _asset(name: str) -> Path:
        return _ASSETS / name

    return _asset
```

pip_audit/_cli.py
```diff
@@ -330,8 +331,14 @@ def _dep_source_from_project_path(
     project_path: Path, resolver: ResolveLibResolver, state: AuditState
 ) -> DependencySource:  # pragma: no cover
     # Check for a `pyproject.toml`
```
woodruffw, line 333: Looks like this comment needs to be moved down/removed entirely (it's a little redundant).

test/dependency_source/test_poetry.py
orsinium: I could. But why? This one is short, type-safe, and not coupled on conftest.py. I love good autocomplete.

pip_audit/_dependency_source/poetry.py
orsinium: removed both ✅

README.md
orsinium: fixed ✅

pip_audit/_dependency_source/poetry.py
orsinium: Changed to DependencyFixError, now it is gracefully handled by CLI ✅

pip_audit/_cli.py
orsinium: removed ✅

test/dependency_source/test_poetry.py
woodruffw: My main reason is potential reuse: we have a lot of other tests (particularly the requirements file ones) that would benefit from being broken out into stand-alone assets, so having a centralized "asset" fixture would keep things unified. 

Separately (and I admit I have no idea if this is "good" practice or not): my understanding is that `pytest` conceptually encourages each test module to contain nothing _except_ test code, with `conftest.py` being the idiomatic place to put any scaffolding.

woodruffw (commented):
LGTM structurally. This PR changes a user-facing component, so we'll need a CHANGELOG entry -- could you add that @orsinium? 🙂

di:
I think merging this is probably blocked on making a determination on whether this tool should support Poetry in any capacity in the long-term. The original/current intention that this will become part of `pip`, in which case having support for Poetry doesn't make much sense.

That discussion should probably happen in #84 before we do any more work on this or merge it.

test/dependency_source/test_poetry.py
orsinium: > we have a lot of other tests (particularly the requirements file ones) that would benefit from being broken out into stand-alone assets

YAGNI ;)

> Separately (and I admit I have no idea if this is "good" practice or not): my understanding is that pytest conceptually encourages each test module to contain nothing except test code

Never heard of that. Test files can contain fixtures, and from my experience, they should when the fixture is used only in that file. Having all fixtures in `conftest.py` brings unnecessary coupling, makes it harder to track which fixtures aren't needed anymore, and that file being a typical example of "God class" (in this case, God module) can grow to some crazy numbers (last month, I decoupled 10k lines conftest.py). And even if we assume "keep all fixtures in conftest.py", it's not a fixture. It's literally a constant.

Having a constant has direct benefits: it's shorter, type-safe, closer to the code (easier to unravel the test code), and makes the test file self-contained. If none of that persuades you, please, just say it, and I'll update the PR. I want the code I contribute to be nice and readable, but the final word is always on the direct maintainers.

PhorstenkampFuzzy:
Isn't this OBE with poetry2? Poetry has a compliant dependency syntax for all direct dependencies.

woodruffw:
Sorry, what does OBE mean?

PhorstenkampFuzzy:
Overcome by events

woodruffw:
Possibly -- my understanding is that Poetry still has its own lock files etc, but I haven't followed things there closely. If Poetry now matches other tools in terms of conforming to the expected shape of pyproject.toml and the new PEP 751 lockfile format, then perhaps this is now obsolete.

PR #404, opened 2022-11-05
Add VulnerabilityResult.published field

orsinium:
Add a new field `published` into `VulnerabilityResult` dataclass. The field holds information about when the vulnerability was published. That information is not used in the CLI or anywhere else yet, only emitted (from both PyPA and OSV sources).

Motivation: in a follow-up PR, I'm going to introduce a new CLI flag for filtering out only advisories published in a specific time window. The motivation for that, in turn, is to be able to have multiple CI jobs in user projects, one that warns about new vulnerabilities and another that fails for all vulnerabilities that weren't resolved in time. But that's another story. For now, just hold that information but don't act on it :)

### DIFF
--- a/pip_audit/_service/interface.py
+++ b/pip_audit/_service/interface.py
@@ -6,8 +6,9 @@ of vulnerability information for fully resolved Python packages.
 from __future__ import annotations
 
 from abc import ABC, abstractmethod
-from dataclasses import dataclass, field
-from typing import Any, Dict, Iterator, List, NewType, Set, Tuple
+from dataclasses import dataclass, field, replace
+from datetime import datetime
+from typing import Any, Iterator, NewType
 
 from packaging.utils import canonicalize_name
 from packaging.version import Version
@@ -58,7 +59,7 @@ class ResolvedDependency(Dependency):
     """
 
     version: Version
-    hashes: Dict[str, List[str]] = field(default_factory=dict, hash=False)
+    hashes: dict[str, list[str]] = field(default_factory=dict, hash=False)
 
 
 @dataclass(frozen=True)
@@ -87,14 +88,19 @@ class VulnerabilityResult:
     A human-readable description of the vulnerability.
     """
 
-    fix_versions: List[Version]
+    fix_versions: list[Version]
     """
     A list of versions that can be upgraded to that resolve the vulnerability.
     """
 
-    aliases: Set[str]
+    aliases: set[str]
     """
     A set of aliases (alternative identifiers) for this result.
+    """
+
+    published: datetime | None = None
+    """
+    When the vulnerability was first published.
     """
 
     def alias_of(self, other: VulnerabilityResult) -> bool:
@@ -112,11 +118,10 @@ class VulnerabilityResult:
         """
 
         # Our own ID should never occur in the alias set.
-        return VulnerabilityResult(
-            self.id, self.description, self.fix_versions, self.aliases | other.aliases - {self.id}
-        )
+        aliases = self.aliases | other.aliases - {self.id}
+        return replace(self, aliases=aliases)
 
-    def has_any_id(self, ids: Set[str]) -> bool:
+    def has_any_id(self, ids: set[str]) -> bool:
         """
         Returns whether ids intersects with {id} | aliases.
         """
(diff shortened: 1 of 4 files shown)


pip_audit/_service/interface.py
```diff
@@ -87,16 +88,21 @@ class VulnerabilityResult:
...
     A list of versions that can be upgraded to that resolve the vulnerability.
     """
 
-    aliases: Set[str]
+    aliases: set[str]
     """
     A set of aliases (alternative identifiers) for this result.
     """
 
+    published: datetime | None = None
```
woodruffw, line 101: Does this work in Python 3.7? I know that PEP 585 (`dict[...]` instead of `Dict[...]`) does with `future`, but PEP 604 doesn't mention similar support.

pip_audit/_service/interface.py
```diff
@@ -150,6 +155,12 @@ def query_all(
         for spec in specs:
             yield self.query(spec)
 
+    @staticmethod
+    def _parse_rfs3339(dt: str | None) -> datetime | None:
```
woodruffw, line 159: Typo, I think:

```suggestion
    def _parse_rfc3339(dt: str | None) -> datetime | None:
```

woodruffw (commented):
Some tiny nits, but LGTM overall! Thanks!

woodruffw:
Thanks for adding this field! This will be a nice addition, when plumbed through the different output formats.

> Motivation: in a follow-up PR, I'm going to introduce a new CLI flag for filtering out only advisories published in a specific time window.

Would you mind opening a discussion issue for this feature first? So far we've tried to limit the number of "filter-style" CLI options we provide, under the reasoning that users who want filtering should use `--format=json` (or another machine-readable format) and do filtering as a post-processing step.

pip_audit/_service/osv.py
```diff
@@ -110,10 +112,10 @@ def query(self, spec: Dependency) -> Tuple[Dependency, List[VulnerabilityResult]
             # formatting in the process).
             description = description.replace("\n", " ")
 
-            aliases = set(vuln.get("aliases", []))
-
             # OSV doesn't mandate this field either. There's very little we
             # can do without it, so we skip any results that are missing it.
+            #
+            # TODO(@orsinium): not true anymore, the field is marked as required in API docs.
```
di, line 118: Let's just revise this comment instead.

di (commented):
LGTM aside from unresolved discussion

pip_audit/_service/osv.py
woodruffw: @orsinium where did you see that this field is marked as required?

Here's what I see in the OSV schema docs (https://ossf.github.io/osv-schema/#id-modified-fields):

> The id and modified fields are required. All other fields are optional, although of course an entry with no other metadata is not particularly useful. (It could potentially stand for a reserved ID with no other public information.)

woodruffw: (It makes sense for `affected` in particular to be optional, for the "reserved ID" use case mentioned in the docs.)

pip_audit/_service/interface.py
orsinium: + PEP 585 is introduced in Python 3.9 and adds support for generic built-ins. 
+ PEP 604 is introduced in Python 3.10 and adds support for union operator on types.
+ PEP 563 is introduced in Python and adds `from __future__ import annotations` making all annotations lazy.
+ Both PEP 585 and PEP 604 can be used in Python 3.7 and 3.9 in combination with PEP 563 because annotations aren't evaluated at runtime.

pip_audit/_service/osv.py
orsinium: In [API docs](https://osv.dev/docs/#tag/api/operation/OSV_QueryAffected) for 200 Response fields:

> affected | Array of objects (osvAffected)
> **Required**. Affected commit ranges and versions.

pip_audit/_service/interface.py
woodruffw: > PEP 563 is introduced in Python 3.7 and adds from __future__ import annotations making all annotations lazy.

TIL, thank you!

pip_audit/_service/osv.py
woodruffw: This is a funny case: OSV's own API might alway return the `affected` key, but the OSV specification says that it's optional. I'm going to open an upstream issue to get clarification here.

woodruffw: Ref: https://github.com/google/osv.dev/issues/810

woodruffw (commented):
LGTM overall, we should probably clarify the OSV schema question but I'm okay with letting that float if you are @di 🙂

di (approved):

woodruffw:
Thanks again @orsinium!

PR #414, opened 2022-11-22
pip_audit, test: modernize type hints

### LINKED ISSUE #405: Modernize our type hints
As @orsinium pointed out in #404, we can use more modern type hint styles without needing to increase our minimum Python version.

In particular, we can use type unions (`Foo | Bar`) and "intrinsic" types (`dict[...]` instead of `Dict[...]`).

woodruffw:
Prefer "intrinsic" hints and bar-style unions over `typing.*`

Closes #405.

### DIFF
--- a/pip_audit/_audit.py
+++ b/pip_audit/_audit.py
@@ -1,10 +1,11 @@
 """
 Core auditing APIs.
 """
+from __future__ import annotations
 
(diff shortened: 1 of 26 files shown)


tetsuo-cpp (approved):
LGTM!

PR #420, opened 2022-11-28
Improve our documentation of `--ignore-vuln`

### LINKED ISSUE #419: Ability to ignore multiple vulnerabilities
I want to ignore multiple vulnerabilities in my project: 

We are using pyspark=3.2.1 as there is no newer version available on Azure at the moment. If we use `pip-audit` as a pre-commit hook, all of our builds keep failing because there are vulnerabilities for pyspark=3.2.1 - which we need to ignore. 

As far as I figured out, I can only pass single vuln-ids.

Ability to pass a list of vulnerability IDs to  `pip-audit --ignore-vuln`

e.g.:  `pip-audit --ignore-vuln ['GHSA-4x9r-j582-cgr8', 'PYSEC-2022-42976']`

woodruffw:
cc @CribberSix: let us know if this helps to clarify how the option can be used!

Closes #419.

### DIFF
--- a/README.md
+++ b/README.md
@@ -205,7 +205,8 @@ optional arguments:
   -o FILE, --output FILE
                         output results to the given file (default: None)
   --ignore-vuln ID      ignore a specific vulnerability by its vulnerability
-                        ID (default: [])
+                        ID; this option can be used multiple times (default:
+                        [])
 ```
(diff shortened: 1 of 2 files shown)


di (approved):

CribberSix:
This solution is great, thanks! :)

PR #451, opened 2022-12-27
pip_audit, test: warn on Python path confusion

### LINKED ISSUE #450: Running globally installed pip-audit from within a virtualenv produces unintuitive results
## Bug description

If you have a globally installed pip-audit (e.g. from `brew install pip-audit`), and then run it inside a virtualenv environment, the results are confusing.

```
(.venv) ~/p/cryptography ❯❯❯ pip-audit 
Found 1 known vulnerability in 1 package
Name       Version ID                  Fix Versions
---------- ------- ------------------- ------------
setuptools 65.5.0  GHSA-r9hx-vwmv-q579 65.5.1
(.venv) ~/p/cryptography ❯❯❯ pip list | grep setuptools
setuptools           65.6.3
setuptools-rust      1.5.2
(.venv) ~/p/cryptography ❯❯❯ which pip-audit
/opt/homebrew/bin/pip-audit
```

## Reproduction steps

1. `brew install pip-audit`
2. `python -m virtualenv venv`
3. `venv/bin/activate`
4. `pip-audit`

## Expected behavior

Either it scans the packages installed inside the virtual environment, or at least emits a warning or an error.

## Platform information

* OS name and version: macOS Ventura 13.1
* `pip-audit` version (`pip-audit -V`): pip-audit 2.4.10
* Python version (`python -V` or `python3 -V`): Python 3.10.9
* `pip` version (`pip -V` or `pip3 -V`): pip 22.3.1 from /Users/alex_gaynor/projects/cryptography/.venv/lib/python3.10/site-packages/pip (python 3.10)

woodruffw:
Closes #450.

### DIFF
--- a/pip_audit/_dependency_source/pip.py
+++ b/pip_audit/_dependency_source/pip.py
@@ -5,7 +5,6 @@ by `pip-api`.
 
 import logging
 import os
-import shutil
 import subprocess
 import sys
 from pathlib import Path
--- a/pip_audit/_dependency_source/pip.py
+++ b/pip_audit/_dependency_source/pip.py
@@ -75,19 +75,32 @@ class PipSource(DependencySource):
         # a loaded virtual environment, since `pip-audit`'s `sys.executable` will be the global
         # Python and not the virtual environment's Python.
         #
-        # Also note that we don't call `Path.resolve()` on the paths below -- doing so would
-        # follow symlinks to e.g. a `pyenv` managed Python, which in turn would pierce
-        # the virtual environment and render the comparison moot.
-        effective_python = Path(os.environ.get("PIPAPI_PYTHON_LOCATION", sys.executable))
-        path_python = Path(shutil.which("python") or sys.executable)
-        if effective_python != path_python:
+        # To check for this, we check whether the Python that `pip_api` plans to use
+        # matches the active virtual environment's prefix. We do this instead of comparing
+        # against the $PATH-prioritized Python because that might be the same "effective"
+        # Python but with a different symlink (e.g. `<path>/python{,3,3.7}`). We *could*
+        # handle that case by resolving the symlinks, but that would then piece the
+        # virtual environment that we're attempting to detect.
+        effective_python = os.environ.get("PIPAPI_PYTHON_LOCATION", sys.executable)
+        venv_prefix = os.getenv("VIRTUAL_ENV")
+        if venv_prefix is not None and not effective_python.startswith(venv_prefix):
             logger.warning(
                 f"pip-audit will run pip against {effective_python}, but you have "
-                f"{path_python} in your PATH (usually indicating a virtual environment). "
-                "This may result in unintuitive audits, since your local environment "
-                "will not be audited. You can forcefully override this behavior by setting "
-                "PIPAPI_PYTHON_LOCATION to the location of your local Python interpreter."
+                f"a virtual environment loaded at {venv_prefix}. This may result in "
+                "unintuitive audits, since your local environment will not be audited. "
+                "You can forcefully override this behavior by setting PIPAPI_PYTHON_LOCATION "
+                "to the location of your virtual environment's Python interpreter."
             )
+
+        # path_python = Path(shutil.which("python") or sys.executable)
+        # if effective_python != path_python:
+        #     logger.warning(
+        #         f"pip-audit will run pip against {effective_python}, but you have "
+        #         f"{path_python} in your PATH (usually indicating a virtual environment). "
+        #         "This may result in unintuitive audits, since your local environment "
+        #         "will not be audited. You can forcefully override this behavior by setting "
+        #         "PIPAPI_PYTHON_LOCATION to the location of your local Python interpreter."
+        #     )
 
         if _PIP_VERSION < _MINIMUM_RELIABLE_PIP_VERSION:
             logger.warning(
--- a/pip_audit/_dependency_source/pip.py
+++ b/pip_audit/_dependency_source/pip.py
@@ -92,16 +92,6 @@ class PipSource(DependencySource):
                 "to the location of your virtual environment's Python interpreter."
             )
 
-        # path_python = Path(shutil.which("python") or sys.executable)
-        # if effective_python != path_python:
-        #     logger.warning(
-        #         f"pip-audit will run pip against {effective_python}, but you have "
-        #         f"{path_python} in your PATH (usually indicating a virtual environment). "
-        #         "This may result in unintuitive audits, since your local environment "
-        #         "will not be audited. You can forcefully override this behavior by setting "
-        #         "PIPAPI_PYTHON_LOCATION to the location of your local Python interpreter."
-        #     )
-
         if _PIP_VERSION < _MINIMUM_RELIABLE_PIP_VERSION:
             logger.warning(
                 f"pip {_PIP_VERSION} is very old, and may not provide reliable "
--- a/pip_audit/_dependency_source/pip.py
+++ b/pip_audit/_dependency_source/pip.py
@@ -4,6 +4,8 @@ by `pip-api`.
 """
 
 import logging
+import os
+import shutil
 import subprocess
 import sys
 from pathlib import Path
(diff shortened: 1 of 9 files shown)


woodruffw:
NB: This doesn't work quite as expected yet:

```
$ pip-audit
WARNING:pip_audit._dependency_source.pip:pip-audit will run pip against /Users/william/.pyenv/versions/3.11.0/bin/python3.11, but you have /Users/william/.pyenv/versions/3.11.0/bin/python in your PATH (usually indicating a virtual environment). This may result in unintuitive audits, since your local environment will not be audited. You can forcefully override this behavior by setting PIPAPI_PYTHON_LOCATION to the location of your local Python interpreter.
Found 2 known vulnerabilities in 2 packages
Name       Version   ID                  Fix Versions
---------- --------- ------------------- ------------
certifi    2022.9.24 GHSA-43fp-rhv2-5gv8 2022.12.7
setuptools 65.5.0    GHSA-r9hx-vwmv-q579 65.5.1
```

(In this case it can't see that the two global versions are actually the same.)

woodruffw:
I can induce the expected warning, however:

```
$ ~/.pyenv/versions/3.11.0/bin/pip-audit
WARNING:pip_audit._dependency_source.pip:pip-audit will run pip against /Users/william/.pyenv/versions/3.11.0/bin/python3.11, but you have /Users/william/devel/pip-audit/env/bin/python in your PATH (usually indicating a virtual environment). This may result in unintuitive audits, since your local environment will not be audited. You can forcefully override this behavior by setting PIPAPI_PYTHON_LOCATION to the location of your local Python interpreter.
Found 2 known vulnerabilities in 2 packages
Name       Version   ID                  Fix Versions
---------- --------- ------------------- ------------
certifi    2022.9.24 GHSA-43fp-rhv2-5gv8 2022.12.7
setuptools 65.5.0    GHSA-r9hx-vwmv-q579 65.5.1
```

woodruffw:
[7f943b5](https://github.com/pypa/pip-audit/pull/451/commits/7f943b5d3d996ed328e46862026867769000251b) should be a better approach -- rather than checking whether `sys.executable` diverges from the `$PATH` Python, we check whether we're in a virtual environment but are executing with a `sys.executable` outside of that environment.

woodruffw:
Example output, formatted for readability (global `pip-audit`, local virtual environment):

```
$ ~/.pyenv/versions/3.11.0/bin/pip-audit 
WARNING:pip_audit._dependency_source.pip:pip-audit will run pip against
/Users/william/.pyenv/versions/3.11.0/bin/python3.11, but you have a virtual environment 
loaded at /Users/william/devel/pip-audit/env. This may result in unintuitive audits, since 
your local environment will not be audited. You can forcefully override this behavior by 
setting PIPAPI_PYTHON_LOCATION to the location of your local Python interpreter.

Found 2 known vulnerabilities in 2 packages
Name       Version   ID                  Fix Versions
---------- --------- ------------------- ------------
certifi    2022.9.24 GHSA-43fp-rhv2-5gv8 2022.12.7
setuptools 65.5.0    GHSA-r9hx-vwmv-q579 65.5.1
```

woodruffw:
Thinking about it some more: maybe we should just do what the user expects here (i.e. configure `PIPAPI_PYTHON_LOCATION` for them, based on `VIRTUAL_ENV`).

We could still emit a warning, but this would avoid users having to do their own manual workaround and would probably match ordinary user expectations (similar to how `mypy` or any other tool can be installed globally but used within a virtual environment).

tetsuo-cpp:
> Thinking about it some more: maybe we should just do what the user expects here (i.e. configure `PIPAPI_PYTHON_LOCATION` for them, based on `VIRTUAL_ENV`).
> 
> We could still emit a warning, but this would avoid users having to do their own manual workaround and would probably match ordinary user expectations (similar to how `mypy` or any other tool can be installed globally but used within a virtual environment).

Seems reasonable to me. If you find a file at `${VIRTUAL_ENV}/bin/python`, you can use that.

pip_audit/_dependency_source/pip.py
```diff
@@ -83,13 +83,17 @@ def __init__(
...
-                "unintuitive audits, since your local environment will not be audited. "
-                "You can forcefully override this behavior by setting PIPAPI_PYTHON_LOCATION "
-                "to the location of your virtual environment's Python interpreter."
-            )
+            venv_python = Path(venv_prefix + "/bin/python3")
+            if venv_python.exists():
+                os.environ["PIPAPI_PYTHON_LOCATION"] = str(venv_python)
+            else:
+                logger.warning(
+                    f"pip-audit will run pip against {effective_python}, but you have "
```
tetsuo-cpp, line 91: I thought about editing this to include the fact that we tried to look for a symlink in `bin/` but wasn't sure how to phrase it as this is already getting complicated. I don't think it's all that important to communicate to the user since the corrective action doesn't change.

tetsuo-cpp: We could also maybe just fail here?

woodruffw: Yeah, I think just failing is probably appropriate -- the lack of a `bin` here suggests that the user intended the load a virtual environment (based on the env setting) but somehow ended up with a broken on, so we shouldn't try to guess at the correct behavior.

woodruffw:
Hmm, there's another issue here: this check:

```
        if _PIP_VERSION < _MINIMUM_RELIABLE_PIP_VERSION:
            logger.warning(
                f"pip {_PIP_VERSION} is very old, and may not provide reliable "
                "dependency information! You are STRONGLY encouraged to upgrade to a "
                "newer version of pip."
            )
```

...doesn't actually do what we (or the user) expect, since `_PIP_VERSION` is just `pip_api.PIP_VERSION` and is therefore loaded eagerly on `pip_api`'s import. So it doesn't see the updated `PIPAPI_PYTHON_PATH`.

woodruffw:
Based on the above, I _think_ we just want to go with the initial "warn" approach for now. 

That'll allow us to get a bugfix out, and from there we can work on actually using the active virtual environment by default (since it'll require us to rethink how we use `pip_api` and possibly make some upstream changes).

woodruffw:
For posterity, here's the tweaked approach based on @tetsuo-cpp's work (mostly tweaked to respect `PIPAPI_PYTHON_LOCATION` if the user has already overridden it):

```python
        # NOTE: By default `pip_api` invokes `pip` through `sys.executable`, like so:
        #
        #    {sys.executable} -m pip [args ...]
        #
        # This is the right decision 99% of the time, but it can result in unintuitive audits
        # for users who have installed `pip-audit` globally but are trying to audit
        # a loaded virtual environment, since `pip-audit`'s `sys.executable` will be the global
        # Python and not the virtual environment's Python.
        #
        # Rather than auditing the global environment (based on how `pip-audit` has been
        # installed), we detect the user's intent by checking for an active virtual
        # environment. If there's an active virtual environment and the user hasn't
        # already overriden `PIPAPI_PYTHON_LOCATION`, then we point `pip_api` at the
        # right Python path in that virtual environment.
        effective_python = os.getenv("PIPAPI_PYTHON_LOCATION")
        if effective_python is None:
            effective_python = sys.executable

            venv_prefix = os.getenv("VIRTUAL_ENV")
            if venv_prefix is not None and not effective_python.startswith(venv_prefix):
                venv_python = Path(venv_prefix) / "bin/python"

                if venv_python.exists():
                    os.environ["PIPAPI_PYTHON_LOCATION"] = str(venv_python)
                else:
                    raise PipSourceError(
                        f"pip-audit found a virtual environment at {venv_prefix}, but "
                        "couldn't find a Python interpreter in that virtual environment"
                    )

        if _PIP_VERSION < _MINIMUM_RELIABLE_PIP_VERSION:
            logger.warning(
                f"pip {_PIP_VERSION} is very old, and may not provide reliable "
                "dependency information! You are STRONGLY encouraged to upgrade to a "
                "newer version of pip."
            )
```

tetsuo-cpp:
> Based on the above, I _think_ we just want to go with the initial "warn" approach for now.

Good idea. I completely missed the issue with `PIP_VERSION`. Let's just merge this for now.

tetsuo-cpp (approved):
LGTM!

PR #454, opened 2022-12-28
Include setuptools in upgrade set

woodruffw:
This looks like an oversight on our part: we should keep setuptools upgraded along with `pip` and `wheel`, since the default environment one can be pretty old.


woodruffw:
(This is currently responsible for a bunch of spurious `gh-action-pip-audit` failures -- plenty of environments have a slightly older `setuptools` that is "vulnerable" to a ReDoS.)

tetsuo-cpp (approved):
LGTM!

PR #461, opened 2023-01-02
pip_audit: `ruff` fixes

woodruffw:
I initially thought this was a false positive, but it looks like CPython's documentation on `IOError` is slightly stale.

xref https://github.com/charliermarsh/ruff/issues/1568


tetsuo-cpp (approved):
LGTM!

PR #462, opened 2023-01-04
requirement, resolvelib: Hash package contents to verify requirement file hashes

### LINKED ISSUE #433: Hash checking: respect third party indices
Yes.

Currently if you are using a privately hosted repo and building your own wheels for various projects, you cannot use `pip-audit -s pypi`, while only listing the hashes of your privately hosted wheels.  This leaves us with 2 options.  Adding the upstream digest hash ( that we dont use ) to our hash list for each package. Or using `-s osv`.  

Currently we are using osv. And though I am unsure if this is inferior.  We would like to use the default where we can. 

Ideally - https://github.com/pypa/pip-audit/blob/76e4fa4a0ed005f543701d4b60c02c75c5b539d2/pip_audit/_service/pypi.py#L89

Would have logic have a flag around it to disable hash checking for situations like this.

As mentioned above using `-s osv`

**Additional context**

Example of current failure:
```DEBUG:cachecontrol.controller:Looking up "https://pypi.org/pypi/adal/1.2.2/json" in the cache
DEBUG:cachecontrol.controller:No cache entry available
DEBUG:urllib3.connectionpool:Starting new HTTPS connection (1): pypi.org:443
DEBUG:urllib3.connectionpool:https://pypi.org:443 "GET /pypi/adal/1.2.2/json HTTP/1.1" 200 1128
DEBUG:cachecontrol.controller:Updating cache with response from "https://pypi.org/pypi/adal/1.2.2/json"
DEBUG:cachecontrol.controller:etag object cached for 1209600 seconds
DEBUG:cachecontrol.controller:Caching due to etag
DEBUG:cachecontrol.controller:Looking up "https://pypi.org/pypi/aenum/3.1.0/json" in the cache
DEBUG:cachecontrol.controller:No cache entry available
DEBUG:urllib3.connectionpool:https://pypi.org:443 "GET /pypi/aenum/3.1.0/json HTTP/1.1" 200 2235
DEBUG:cachecontrol.controller:Updating cache with response from "https://pypi.org/pypi/aenum/3.1.0/json"
DEBUG:cachecontrol.controller:etag object cached for 1209600 seconds
DEBUG:cachecontrol.controller:Caching due to etag
DEBUG:cachecontrol.controller:Looking up "https://pypi.org/pypi/aiobotocore/1.2.2/json" in the cache
DEBUG:cachecontrol.controller:No cache entry available
DEBUG:urllib3.connectionpool:https://pypi.org:443 "GET /pypi/aiobotocore/1.2.2/json HTTP/1.1" 200 5390
DEBUG:cachecontrol.controller:Updating cache with response from "https://pypi.org/pypi/aiobotocore/1.2.2/json"
DEBUG:cachecontrol.controller:etag object cached for 1209600 seconds
DEBUG:cachecontrol.controller:Caching due to etag
  File "/Users/carl.gill/.pyenv/versions/3.7.3/bin/pip-audit", line 8, in <module>
    sys.exit(audit())
  File "/Users/carl.gill/.pyenv/versions/3.7.3/lib/python3.7/site-packages/pip_audit/_cli.py", line 439, in audit
    for (spec, vulns) in auditor.audit(source):
  File "/Users/carl.gill/.pyenv/versions/3.7.3/lib/python3.7/site-packages/pip_audit/_audit.py", line 67, in audit
    for dep, vulns in self._service.query_all(specs):
  File "/Users/carl.gill/.pyenv/versions/3.7.3/lib/python3.7/site-packages/pip_audit/_service/interface.py", line 156, in query_all
    yield self.query(spec)
  File "/Users/carl.gill/.pyenv/versions/3.7.3/lib/python3.7/site-packages/pip_audit/_service/pypi.py", line 101, in query
    f"Mismatched hash for {spec.canonical_name} ({spec.version}): listed "
pip_audit._service.interface.ServiceError: Mismatched hash for aiobotocore (1.2.2): listed 290833e77f3992cf947279106e43f97dc66f72731cf6818d89a171295bda79ba of type sha256 could not be found in PyPI releases```

### PR DESCRIPTION
**Pull request by tetsuo-cpp**
Closes #433

### CODE REVIEW

**Comment by tetsuo-cpp** [comment]
Just pushing for visibility, this is not ready.

In order to get this to work, I've had to push hashed requirements file down the dependency resolution path where we can no longer take advantage of the fact that these files are guaranteed to be fully resolved.

There's some really ugly bits here that I'm not super pleased with. I'll sleep on it and come back to it tomorrow but if anyone has a better design, I'd be glad to hear it.

File: pip_audit/_dependency_source/resolvelib/resolvelib.py [hunk 1]
```diff
@@ -72,6 +77,9 @@ def resolve(self, req: Requirement) -> list[Dependency]:
                     SkippedDependency(name=req.name, skip_reason="requirement marked as editable")
                 ]
 
+        # TODO(alex): Figure out a better way to do this
+        self.provider.req_hashes = req_hashes
```
tetsuo-cpp, line 81: The provider gets set when the resolver is constructed (which needs to happen to construct the requirements file). So we obviously can't have the hash information from the requirements file at the point of construction.

We could just create a new `provider` for each `resolve` call... not sure that's really the right answer though.

pip_audit/_dependency_source/requirement.py
```diff
@@ -306,26 +297,30 @@ def _collect_cached_deps(
...
+            req_hashes = {}
+
+            # If we're requiring hashes, enforce that all requirements are hashed
+            if require_hashes:
+                for req in reqs:
+                    if not req.hash_options:
+                        raise RequirementSourceError(
+                            f"requirement {req.name} does not contain a hash {str(req)}"
+                        )
+                    assert req.name not in req_hashes
```
woodruffw, line 318: Making sure I understand: what is this `assert` doing? It looks like it's checking to make sure we haven't seen this requirement before, but that's probably something that we should factor and ensure we check for all requirements files (given that it's a sign of self-inconsistency in the input).

tetsuo-cpp: Yeah it's asserting that we haven't seen it before. We've already checked for duplicates elsewhere so this should never be the case.

I thought I'd just assert here so that if the code ever changes, we're going to fail loudly here instead of silently overwriting that dict key.

pip_audit/_dependency_source/resolvelib/pypi_provider.py
```diff
@@ -188,12 +195,26 @@ def _get_metadata_for_sdist(self) -> Message:
...
+    def _generate_dist_hashes(self, dist_data: bytes) -> None:
+        """
+        Populate a mapping of hash algorithm names to hashes for the candidate.
+        """
+        if self.name not in self.req_hashes:
+            return
+        hash_algorithms = self.req_hashes[self.name].keys()
+        dist_hashes = {}
+        for alg in hash_algorithms:
+            hasher = hashlib.new(alg, dist_data)
```
woodruffw, line 207: Small thing, but `hashlib.new` actually has subtle behavior we don't necessarily want: it'll dynamically look up *any* digest functions provided by the underlying OpenSSL library, even ones that aren't explicitly supported by `hashlib` itself.

To work around that, we probably want to check `hash_algorithms` against `hashlib.algorithms_guaranteed` first, and warn/error on any identifiers that we don't recognize.

pip_audit/_dependency_source/resolvelib/resolvelib.py
woodruffw: Yeah, that's not ideal. I'll look at this a bit.

IMO, creating a new provider for each resolution isn't the _worst_ idea -- we're reusing HTTP sessions anyways and caching everything, so the provider objects themselves are _relatively_ lightweight (especially in the context of network-dominated execution).

pip_audit/_dependency_source/resolvelib/resolvelib.py
```diff
@@ -82,6 +90,17 @@ def resolve(self, req: Requirement) -> list[Dependency]:
...
+            if req_hashes:
+                for alg, hashes in req_hashes[name].items():
+                    for h in hashes:
+                        if h == candidate.dist_hashes[alg]:
+                            break
+                    else:
+                        raise ResolveLibResolverError(
+                            f"Mismatched hash for {name} ({candidate.version}): listed {h} of type "
+                            f"{alg} could not be found"
+                        )
```
woodruffw, line 103: This kind of boilerplate is definitely a good argument for having `RequirementHashes` be a full model with a `match` or similar method 🙂

pip_audit/_dependency_source/resolvelib/pypi_provider.py
tetsuo-cpp: Good catch. Thanks.

pip_audit/_dependency_source/resolvelib/resolvelib.py
tetsuo-cpp: [698d897](https://github.com/pypa/pip-audit/pull/462/commits/698d897b04186c909c0cd30a97e504f4b7703277)

pip_audit/_dependency_source/resolvelib/resolvelib.py
woodruffw: Thought about this some more, and I think creating a new provider on each resolution is fine (and preferable for now) -- there's no meaningful overhead due to the shared resources, and it avoids a `None` typestate.

tetsuo-cpp: [31d7cff](https://github.com/pypa/pip-audit/pull/462/commits/31d7cffc51e5c29bf40861dd6a6f59ff96c8c900)

pip_audit/_dependency_source/requirement.py
tetsuo-cpp: Sure, I'll just turn that into a check and cover it with the unit tests.

pip_audit/_dependency_source/resolvelib/pypi_provider.py
tetsuo-cpp: [911cd0d](https://github.com/pypa/pip-audit/pull/462/commits/911cd0d2cde1c439de3f5cef43216feeae49d3e9)

pip_audit/_dependency_source/resolvelib/pypi_provider.py
```diff
@@ -95,10 +101,15 @@ def metadata(self) -> Message:
         if self._metadata is None:
             self._state.update_state(f"Fetching metadata for {self.name} ({self.version})")
 
+            response = self._session.get(self.url, timeout=self._timeout)
```
tetsuo-cpp, line 104: I moved this out here so that we can mock out the `get_metadata_for_X` functions to short-circuit dependency resolution without losing the hashes that are generated in `generate_dist_hashes`. It also removes a bit of redundancy.

pip_audit/_dependency_source/resolvelib/resolvelib.py
```diff
@@ -53,12 +60,18 @@ def __init__(
 
         `state` is an `AuditState` to use for state callbacks.
         """
-        self.provider = PyPIProvider(index_urls, timeout, cache_dir, state)
+        self.index_urls = index_urls
+        self.timeout = timeout
+        self.session = caching_session(cache_dir, use_pip=True)
```
tetsuo-cpp, line 65: I'm creating the `session` here so I can patch it in the tests. It's harder to keep it in the `PyPIProvider` because those get created on every `resolve` call.

pip_audit/_dependency_source/resolvelib/pypi_provider.py
woodruffw: Makes sense!

pip_audit/_dependency_source/resolvelib/resolvelib.py
woodruffw: Works for me -- IMO it's worth a comment explaining that we're doing it to make testing a little easier, but `requests` sessions are self-managing so I have no objections 🙂

pip_audit/_dependency_source/requirement.py
woodruffw: > The options are either to remove the assert and just keep it as is.

Yeah, let's go ahead and remove it then -- if we're already enforcing consistency elsewhere and we have unit tests backing that up, then I see no reason to dupe our work 🙂

woodruffw (approved):
LGTM! A few small nits, but overall this looks great!

(We should also make sure to add a `CHANGELOG` entry before merging.)

pip_audit/_dependency_source/requirement.py
tetsuo-cpp: [0701244](https://github.com/pypa/pip-audit/pull/462/commits/0701244bf452d8917455ff2275db3daa3750c258)

pip_audit/_dependency_source/resolvelib/resolvelib.py
tetsuo-cpp: [aae5e80](https://github.com/pypa/pip-audit/pull/462/commits/aae5e80695526e85a1831c5e4b4164e7cf1c256c)

PR #488, opened 2023-01-17
resolvelib: Stop resolving requirements one-by-one

### LINKED ISSUE #463: Stop resolving requirements one-by-one
I suspect that dependency resolution might be subtly wrong. Throughout the codebase, we follow an idiom where we have interfaces that have a function like `foo` to process a single item and then `foo_all` which returns an iterator.

We also apply this idiom in dependency resolution and end up processing requirements one by one: https://github.com/pypa/pip-audit/blob/main/pip_audit/_dependency_source/resolvelib/resolvelib.py#L77. This doesn't seem right to me because if two packages share the same sub-dependency, both packages requirement for that sub-dependency are going to be used to figure out which version to pick. I think that we need to feed the full list of requirements into the resolver for this it work correctly.

CC: @woodruffw @di

tetsuo-cpp:
Closes #463


tetsuo-cpp:
Still need to fix tests + address a few things.

pip_audit/_dependency_source/requirement.py
```diff
@@ -205,9 +205,12 @@ def _fix_file(self, filename: Path, fix_version: ResolvedFixVersion) -> None:
                         r for r in reqs if isinstance(r, InstallRequirement)
                     ]
                     origin_reqs: set[Requirement] = set()
-                    for req, dep in self._collect_cached_deps(filename, list(installed_reqs)):
+                    for dep in self._collect_cached_deps(filename, list(installed_reqs)):
+                        # NOTE(alex): Now we can't map requirements to dependencies, this is broken.
```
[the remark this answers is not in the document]
woodruffw: Yeah, this might be tough. I'll take a look through the `resolvelib` APIs as well, but stashing it in the candidates would be a reasonable approach IMO (since `resolvelib` is generic over the candidate type for purposes like this).

pip_audit/_dependency_source/resolvelib/resolvelib.py
```diff
@@ -92,3 +105,48 @@ class ResolveLibResolverError(DependencyResolverError):
...
+    dependee_map = {}
+    for c in candidates:
+        for dep in c.dependencies:
+            if dep not in dependee_map:
+                dependee_map[dep] = []
+            dependee_map[dep].extend(c.reqs)
+    return dependee_map
+
+
+def _find_origin_reqs(
```
tetsuo-cpp, line 124: @woodruffw How does this approach look to you? I store the requirements for each candidate. Then I make a mapping to go from each candidate's dependencies back to those requirements. That way, I can work backwards from each candidate until I get to a top-level requirement.

tetsuo-cpp:
This is what it looks like when I audit the same requirements file from earlier:
```console
(env) tetsuo@Alexs-MacBook-Pro pip-audit % pip-audit -v -r dependency-resolution-test.txt
DEBUG:pip_audit._cli:parsed arguments: Namespace(local=False, requirements=[<_io.TextIOWrapper name='dependency-resolution-test.txt' mode='r' encoding='UTF-8'>], project_path=None, format=<OutputFormatChoice.Columns: 'columns'>, vulnerability_service=<VulnerabilityServiceChoice.Pypi: 'pypi'>, dry_run=False, strict=False, desc=<VulnerabilityDescriptionChoice.Auto: 'auto'>, cache_dir=None, progress_spinner=<ProgressSpinnerChoice.On: 'on'>, timeout=15, paths=[], verbose=1, fix=False, require_hashes=False, index_url='https://pypi.org/simple/', extra_index_urls=[], skip_editable=False, no_deps=False, output=PosixPath('stdout'), ignore_vulns=[])
DEBUG:pip_audit._cli:Auditing flask (1.0)
DEBUG:pip_audit._cli:Auditing jinja2 (2.10)
DEBUG:pip_audit._cli:Auditing itsdangerous (2.1.2)
DEBUG:pip_audit._cli:Auditing click (8.1.3)
DEBUG:pip_audit._cli:Auditing werkzeug (2.2.2)
DEBUG:pip_audit._cli:Auditing markupsafe (2.1.2)
Found 2 known vulnerabilities in 1 package
Name   Version ID             Fix Versions
------ ------- -------------- ------------
jinja2 2.10    PYSEC-2021-66  2.11.3
jinja2 2.10    PYSEC-2019-217 2.10.1
```

So it audits Jinja2 once and uses the correct version.

pip_audit/_dependency_source/resolvelib/resolvelib.py
woodruffw: Looks reasonable to me!

My only request would be that we give it a slightly more descriptive name 🙂 -- it's a reverse-dependency lookup on the first level of the dependency tree, so maybe `_dependees_for_candidate` or similar?

tetsuo-cpp:
This change requires #499.

pip_audit/_dependency_source/resolvelib/pypi_provider.py
```diff
@@ -418,35 +425,43 @@ def find_matches(
...
+                    and candidate.name == identifier
+                ],
+                key=attrgetter("version", "is_wheel"),
+                reverse=True,
+            )
+        except PyPINotFoundError as e:
+            skip_reason = str(e)
+            logger.debug(skip_reason)
+            self.skip_deps.append(SkippedDependency(name=identifier, skip_reason=skip_reason))
+            return
```
tetsuo-cpp, line 464: @woodruffw I just found a serious problem here as I was fixing unit tests. The fact that we record the skipped candidate is fine, but then we proceed to return 0 candidates for a given requirement. This causes the resolver to raise a `ResolutionImpossible`.

In the past, we just skipped the top-level requirement and then try resolving the next one (since it was doing one at a time). But the resolver is "all or nothing" so we can't really get partial results from it. And once we get into `find_matches`, I don't think there's a way to backtrack and say "nevermind, just forget about that requirement I said I needed earlier". I'm at a bit of a loss figuring out how to fix this.

pip_audit/_dependency_source/resolvelib/resolvelib.py
tetsuo-cpp: [5eeef40](https://github.com/pypa/pip-audit/pull/488/commits/5eeef40aff20073c1721bfe9712d864eae5e4342)

pip_audit/_dependency_source/resolvelib/pypi_provider.py
woodruffw: Yeah, this is reminding me why I thought "one at a time" made sense despite not accurately reflecting the actual dependency resolution process 😅 

IMO a dummy candidate here is fine, but we should probably tread carefully -- it'd be good to have it be a separate type entirely similarly to how we handle `SkippedDependency` (`SkippedCandidate` maybe?).

tetsuo-cpp:
Oops, not quite ready yet. I'm missing a bit of coverage. I'll try to do that later tonight.

One caveat here. I'm thinking this might not be 100% correct yet because it still processes requirements files one-by-one. I'm not sure how `pip` works in this situation but perhaps it gathers requirements form all files and resolves them all at once? I'll need to check.

This patch gets us most of the way there but there still needs some work to fix up the caching to not work on a file basis.

This fixes most cases so I'm proposing to get this in first and cut a release before fixing the multiple file case.

woodruffw:
> One caveat here. I'm thinking this might not be 100% correct yet because it still processes requirements files one-by-one. I'm not sure how `pip` works in this situation but perhaps it gathers requirements form all files and resolves them all at once? I'll need to check.

Yeah, I believe it gathers them all and resolves them in a single shot (for exactly the reason you said.) 

> This fixes most cases so I'm proposing to get this in first and cut a release before fixing the multiple file case.

Strong agree! Multiple requirements inputs are probably only used by a small minority of users, so we shouldn't block a fix for them on a larger fix like this one 🙂

pip_audit/_dependency_source/resolvelib/pypi_provider.py
```diff
@@ -465,6 +500,8 @@ def is_satisfied_by(self, requirement: Any, candidate: Any) -> bool:
         """
         See `resolvelib.providers.AbstractProvider.is_satisfied_by`.
         """
+        if isinstance(candidate, SkippedCandidate):
+            return True
```
woodruffw, line 504: Nit: IMO we should have a small comment here and in `get_dependencies` explaining why we do this (`SkippedCandidate` already explains this, so maybe just a ref telling the reader to look at the docstring there?)

woodruffw (approved):
Approach LGTM overall! I'm giving this a prospective approval so that you can merge it when you feel it's ready/when coverage is back up 🙂

PR #495, opened 2023-01-20
docs: update example of using pip-audit with pipenv

redbmk:
`pipenv` introduced a new `pipenv requirements` command that simplifies the process for generating a `requirements.txt` file, which can be piped directly into `pip-audit`


woodruffw:
Excellent! I'm glad we can get rid of another external tool reference; thanks a ton for the change!

woodruffw (approved):

PR #499, opened 2023-01-25
requirement: Fix `--skip-editable` flag

tetsuo-cpp:
This has been broken since we moved away from `pip-api` to `pip-requirements-parser` for parsing. Unfortunately, unit tests didn't pick up on this since we were passing the `pip-api` object directly into the resolver.

### DIFF
--- a/test/dependency_source/test_requirement.py
+++ b/test/dependency_source/test_requirement.py
@@ -169,6 +169,20 @@ def test_requirement_source_non_editable_without_egg_fragment(monkeypatch):
     )
 
 
+@pytest.mark.online
+def test_requirement_source_editable_skip(monkeypatch):
+    source = requirement.RequirementSource(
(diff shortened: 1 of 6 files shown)


test/dependency_source/test_requirement.py
```diff
@@ -169,6 +169,20 @@ def test_requirement_source_non_editable_without_egg_fragment(monkeypatch):
     )
 
 
+@pytest.mark.online
```
woodruffw, line 172: Does this test need to be online? From the body it looks like it shouldn't need network access 🙂

woodruffw (commented):
LGTM mod one test question + needs a CHANGELOG entry, but I can handle that.

woodruffw (approved):

test/dependency_source/test_requirement.py
tetsuo-cpp: Oops! I'll fix this in #488.

PR #505, opened 2023-01-28
pyproject.toml: Don't check for line-length with `ruff` since we're reformatting with `black`

tetsuo-cpp:
I've found this to be mildly annoying when running `make reformat`. Basically, `ruff` complains about lines being too long and I have to break lines manually to get past `ruff` and let `black` run.

`black` also needs to run after `ruff` since `ruff` could make edits that are incompatible with `black` so if we don't run `black` twice, we'd get into a situation where we'd need to run `make reformat` twice to make sure that there aren't any fixes needed.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -65,6 +65,7 @@ lint: $(VENV)/pyvenv.cfg
 .PHONY: reformat
 reformat:
 	. $(VENV_BIN)/activate && \
+		black $(ALL_PY_SRCS) && \
 		ruff --fix $(ALL_PY_SRCS) && \
 		black $(ALL_PY_SRCS) && \
 		isort $(ALL_PY_SRCS)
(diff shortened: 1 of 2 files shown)


woodruffw:
Oh goody, no fixpoints 🙂 

I'm okay with this, although I wonder if we can just disable `ruff`'s line length checks entirely -- `black` should be entirely sufficient, and that would save us from having to run it twice. 

(I'll approve and leave it to your judgement to determine whether than makes sense to do.)

woodruffw (approved):
LGTM, but see comment about another possible approach.

tetsuo-cpp:
> I'm okay with this, although I wonder if we can just disable `ruff`'s line length checks entirely -- `black` should be entirely sufficient, and that would save us from having to run it twice.

Ah, yeah you're right. Let's do that instead.

Issue #478, opened 2023-01-12
Add a config/command line option to treat some downstream errors as warning instead of complete error exit


norg:
**Is your feature request related to a problem? Please describe.**

From time to time we run into issues where `pip-audit` stops working with a traceback or error that is not related to `pip-audit` itself but for reasons in the projects that are audited.

So for example `celery` is not following PEP440 currently for the `requirements.txt`, this results in a traceback when running `pip-audit` since `packaging` is now more strict on that. So until `celery` is changing the file a `pip-audit` run that has a dependency that at one point ends up with an issue, the whole audit breaks.

I would argue that his can happen more often and a user would still like to have a report for the packages that are resolved.

**Describe the solution you'd like**

Ideally the user can configure this and `pip-audit` will proceed with the rest of the audit and report a warning that some parts couldn't be audited due to errors.

The advantage of the config option is, that the default could still be a clear error and exit by `pip-audit` but users who would like to have a partial report have an alternative, especially if the proper fix in the other package might take time or isn't even considered.

**Describe alternatives you've considered**

Reporting the issues to the project is the proper way but until some things are fixed `pip-audit` runs would always fail.

**Additional context**

Example in Celery https://github.com/celery/celery/discussions/8001#discussion-4746963 which is detected by `packaging` https://github.com/pypa/packaging/issues/658

woodruffw:
Thanks for the report! 

To clarify, the problem is requirements specifiers like this, right?

```
pytz (>dev)
```

Could you also post the `pip-audit` traceback you're seeing for that?

Given that we've jumped a bit ahead of `pip` and the rest of the Python ecosystem in terms of using the latest `packaging` version, I agree that it probably makes sense for us to add an option/configuration setting that allows us to warn and skip rather than fail on non-PEP 440 specifiers. 

cc @di for thoughts.

norg:
> To clarify, the problem is requirements specifiers like this, right?
> 
> ```
> pytz (>dev)
> ```
> 

Yep, exactly this was one of the issues. But I remember that for example an older version of `python-ldap` didn't work with the pip in a CI environment and thus the whole run was also killed. So I guess there are several scenarios where an error deeper in the chain result in an overall error/exit that could be treated as a warning and "skipped" with a note.

> Could you also post the `pip-audit` traceback you're seeing for that?
> 

```
Traceback (most recent call last):
  File "/home/andi/.local/lib/python3.10/site-packages/packaging/requirements.py", line 35, in __init__
[...]

PR #507, opened 2023-01-30
Better errors when an invalid requirement is encountered

woodruffw:
See #478: this isn't a full fix, but will hopefully improve the overall quality and debuggability of the errors produced.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,12 @@ All versions prior to 0.0.9 are untracked.
 
 ## [Unreleased]
 
+### Changed
+
+* Improved error messaging when a requirements input or indirect dependency
+  has an invalid (non-PEP 440) requirements specifier
+  ([#507](https://github.com/pypa/pip-audit/pull/507))
+
 ### Fixed
 
 * Fixed an issue where hash checking would fail when using third-party indices
--- a/pip_audit/_dependency_source/__init__.py
+++ b/pip_audit/_dependency_source/__init__.py
@@ -10,6 +10,7 @@ from .interface import (
     DependencySourceError,
     HashMismatchError,
     HashMissingError,
+    InvalidRequirementSpecifier,
     RequirementHashes,
     UnsupportedHashAlgorithm,
 )
@@ -27,6 +28,7 @@ __all__ = [
     "DependencySourceError",
     "HashMismatchError",
     "HashMissingError",
+    "InvalidRequirementSpecifier",
     "PipSource",
     "PipSourceError",
     "PyProjectSource",
--- a/pip_audit/_dependency_source/interface.py
+++ b/pip_audit/_dependency_source/interface.py
@@ -86,6 +86,13 @@ class UnsupportedHashAlgorithm(Exception):
     pass
 
 
+class InvalidRequirementSpecifier(DependencySourceError):
+    """
+    A `DependencySourceError` specialized for the case of a non-PEP 440 requirements
+    specifier.
+    """
+
+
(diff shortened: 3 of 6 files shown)


woodruffw:
Example with a subdependency:

```
$ pip-audit -r <(echo 'celery==4.4.7')
ERROR:pip_audit._cli:celery has invalid requirement 'pytz (>dev)': Expected closing RIGHT_PARENTHESIS
    pytz (>dev)
         ~^
```

The direct dependency error still isn't great:

```
$ pip-audit -r <(echo 'pytz (>dev)')
ERROR:pip_audit._cli:requirement file /dev/fd/63 contains invalid lines: [InvalidRequirementLine(requirement_line=RequirementLine(line_number=1, line='pytz (>dev)', filename=PosixPath('/dev/fd/63')), error_message='Invalid requirement: : Expected closing RIGHT_PARENTHESIS\n    pytz (>dev)\n         ~^')]
```

...so I'll fix that one up as well.

woodruffw:
Well, it's still not fantastic, but this makes it a bit better:

```
$ pip-audit -r <(echo 'pytz (>dev)')
ERROR:pip_audit._cli:requirement file /dev/fd/63 contains invalid specifier at line 1: Invalid requirement: : Expected closing RIGHT_PARENTHESIS
    pytz (>dev)
         ~^
```

norg:
This is already a nice improvement. But it still prevents the further audit of other packages :)
So if you use a `requrements.txt` with the `celery` but also a compliant one, you would still not get any report on that.

woodruffw:
Yeah, I need to do some more thinking about how we should handle that...it's arguably incorrect of us to skip things just because we can't parse them, since there's probably a positive relationship between "old enough to have an invalid specifier" and "more likely to have known vulnerabilities."

di:
...we also _could_ continue supporting legacy versions: see https://github.com/di/packaging_legacy and https://github.com/pypa/packaging/pull/669.

woodruffw:
> ...we also _could_ continue supporting legacy versions: see https://github.com/di/packaging_legacy and [pypa/packaging#669](https://github.com/pypa/packaging/pull/669).

That works for me, although I wonder if we should establish some kind of (weak) policy around these kinds of compatibility concerns. Three ideas:

1. "Forcing function": as it is, `pip-audit` serves as a forcing function for packaging ecosystem changes: we adopt changes to `packaging` and `resolvelib` before `pip` and others do, meaning that we make potentially breaking changes ahead of the rest of the ecosystem. This is good in terms of encouraging users to align with the standards, but makes adoption for entrenched users harder.
2. "Same as": we generally aim for CLI flag compatibility with `pip`, and maybe we should follow a similar policy around standards behavior: we should only adopt major incompatible changes when `pip` rolls them out. In that case, we should do what you suggested and use `packaging_legacy`, etc. to handle `LegacyVersion`s here.
3. "Conservative": we could lag behind `pip` and other tooling, since `pip-audit` _should_ help users with older codebases/dependency trees (as that's where we're especially likely to be useful). This however entails being more conservative than we currently are with e.g. `pip` support, which might be burdensome for a small tail of users (who we'd like to encourage to upgrade anyways).

tetsuo-cpp (approved):
LGTM!

tetsuo-cpp:
As discussed, we can probably do better here. Either way, this is still a good improvement so let's get it in.

woodruffw:
Yep! I'll keep #478 open so we can come up with a more permanent solution.

norg:
Thanks again for working on that :)

PR #523, opened 2023-02-12
treewide: Use `pip install` for dependency resolution

### LINKED ISSUE #522: Figure out how to correctly retrieve dependencies from source distributions
Our current method of getting dependency requirements for a source distribution is to install it in a virtual environment and then query the list of packages with `pip list`. We then construct a series of requirements that look like: `{package}=={version}`. This is overly restrictive and doesn't capture what the actual requirement is. You can see the code that does that [here](https://github.com/pypa/pip-audit/blob/main/pip_audit/_dependency_source/resolvelib/pypi_provider.py#L221).

For example, a package `foo` might require `bar>=1.0`. When we install into a virtual environment we'll probably install the latest version (let's say that's `bar-3.0`). We'll assume that the requirement is `bar==3.0`. If we later get a package that requires `bar==2.0`, then we'll throw an impossible resolution because we can't satisfy both of these requirements. In reality, we should have just selected `bar-2.0`.

I think that in order to continue with this current approach where we bundle our own resolver, we need to somehow teach `pip-audit` how to find requirements for arbitrary source distributions, similar to what `pip` figures out when you do `pip install .`. This seems difficult as `pip` probably doesn't expose any of this in a way that we can consume as a library.

The other possibility is that we remove the resolver entirely and just lean on `pip` to do everything and do `pip install -r` (and perhaps the new `--dry-run` and `--report` flags) and `pip list` inside a virtual environment.

CC: @woodruffw @jleightcap @di

tetsuo-cpp:
Closes #522


tetsuo-cpp:
- [ ] Figure out how to display `pip` output without breaking the spinner.
- [ ] Fix unit tests.

tetsuo-cpp:
@woodruffw This seems to work ok. I think the only question is how to `pip` output to play nicely with our progress spinner.

woodruffw:
Yep, looks like what I was thinking. Thanks!

> I think the only question is how to pip output to play nicely with our progress spinner.

We depend on `rich` now, so maybe we can leverage `rich`'s ability to create sub-panels to render `pip`'s output underneath the progress spinner. I believe @ESultanik did an experiment with that recently.

tetsuo-cpp:
> I'm happy to explain how I did it, or to help out. `rich` only allows one "live" renderable at a time, which includes the `Progress` object, so I needed to create a custom `Progress` subclass that yields a `Panel` renderable when needed.

Thanks! I'll have a look at `apt-trace` and give it a try. I might ping you if I have questions.

tetsuo-cpp:
Ok, I think I've got something working. Unfortunately, I had to copy part of the `Status` code out of `rich` since it was mostly wedded to the default spinner implementation. 

@di What do you think of this? Here's a recording of how it looks: https://asciinema.org/a/559680
We can put a subtitle on the panel but I wasn't sure what to put since the actual status message says it all.

pip_audit/_state.py
```diff
@@ -98,6 +102,100 @@ def finalize(self) -> None:
...
+class StatusLog:
+    """
+    Displays a status indicator with an optional log panel to display logs
+    for external processes.
+
+    This code is based off of Rich's `Status` component:
+        https://github.com/Textualize/rich/blob/master/rich/status.py
+    """
+
+    LOG_PANEL_HEIGHT = 10
```
tetsuo-cpp, line 114: This isn't perfect. The actual height of the panel is 10 characters and I'm getting the last 10 lines of the log. However, the panel doesn't display the last 10 lines if some of the lines are long enough to wrap in the panel. I'll have a look online and see what I can do about that.

woodruffw: Making sure I understand: does that mean that the panel will "skip" some lines, since some long lines wrap and effectively push the lines that follow out of sight?

woodruffw:
> Ok, I think I've got something working. Unfortunately, I had to copy part of the Status code out of rich since it was mostly wedded to the default spinner implementation.

Fine by me personally! One of `rich`'s strengths is the component model, so partially duplicating an existing model is perfectly fine.

pip_audit/_virtual_env.py
```diff
@@ -106,7 +106,7 @@ def post_setup(self, context: SimpleNamespace) -> None:
             *self._install_args,
         ]
         try:
-            run(package_install_cmd, state=self._state)
+            run(package_install_cmd, log_stdout=True, state=self._state)
```
woodruffw, line 109: Let's also try and do this with `--dry-run --report`: there's a chance that might be faster than doing a full "normal" install, and the JSON it produces should be easier to consume.

woodruffw: Ref: https://github.com/pypa/pip/pull/10771

pip_audit/_state.py
tetsuo-cpp: Yeah that's exactly right. We might not necessarily see the last few lines in the log because of this.

woodruffw: Yeah, I think that's an acceptable trade-off for now. Let's leave a `NOTE`, but otherwise IMO we shouldn't consider this a blocking issue.

pip_audit/_virtual_env.py
tetsuo-cpp: [6bd5a21](https://github.com/pypa/pip-audit/pull/523/commits/6bd5a211d7af322a6e1aeea6f10de69ae27ac278)

pip_audit/_state.py
tetsuo-cpp: [3196a16](https://github.com/pypa/pip-audit/pull/523/commits/3196a16c67fcac0475c0b8fb03eaf3d24d0802b9)

woodruffw:
Screenshot 2023-02-17 at 10 30 46 AM

Love to see it 🙂

pip_audit/_dependency_source/requirement.py
```diff
@@ -240,24 +199,16 @@ def _fix_file(self, filename: Path, fix_version: ResolvedFixVersion) -> None:
...
-                        f"{filename}: {fix_version.dep.canonical_name}"
-                    )
-                    dependee_reqs_formatted = ",".join(
-                        [
-                            str(req)
-                            for req in sorted(list(req_dep.dependee_reqs), key=lambda x: x.name)
-                        ]
-                    )
-                    print(
-                        f"    # pip-audit: subdependency fixed via {dependee_reqs_formatted}",
```
tetsuo-cpp, line 257: Unfortunately, I don't think we can support this annotation anymore now that we don't have raw access to the resolver.

pip_audit/_dependency_source/requirement.py
```diff
@@ -86,81 +83,45 @@ def collect(self) -> Iterator[Dependency]:
...
+                raise InvalidRequirementSpecifier(
+                    f"requirement file {filename} contains invalid specifier at "
+                    f"line {invalid.line_number}: {invalid.error_message}"
+                )
+
+            # If one or more requirements have a hash, this implies `--require-hashes`.
+            require_hashes = require_hashes or any(req.hash_options for req in rf.requirements)
+            reqs.extend(rf.requirements)
+
+        # If the user has supplied `--no-deps` or there are hashed requirements, we should assume
```
tetsuo-cpp, line 103: This section needs to be reviewed carefully.

One weird side effect of this approach is that by avoiding `pip`, we can get different results when auditing fully resolved requirements files versus invoking `pip` since there are now 2 completely separate modes of operation. Our code to handle pre-resolved dependencies has to skip certain types of requirements (URL requirements and editable) whereas the unresolved dependency flow where we invoke `pip` will handle these without a problem. So it's possible that just by adding hashes to a requirements file, users may get a less comprehensive audit.

I think there's a discussion to be had about whether it makes sense to keep this "preresolved" optimisation since `pip install` could be more efficient than our previous homegrown resolver. The new method won't have the problem we used to have where we potentially install a sdist for each candidate just to figure out its dependencies.

tetsuo-cpp:
@woodruffw @di 

There were a lot of details to get right here, but I think this should be ready now. Please take a look at the comments I've left above. I think we should proceed with this design but a few catches to be aware of.

pip_audit/_dependency_source/requirement.py
```diff
@@ -273,20 +224,49 @@ def _recover_files(self, tmp_files: list[IO[str]]) -> None:
...
+                # For URL requirements that don't have an egg fragment that lists the
+                # package name and version, `pip-requirements-parser` won't attach a
+                # `Requirement` object to the `InstallRequirement`.
+                #
+                # In this case, we can't audit the dependency so we should signal to the
+                # caller that we're skipping it.
+                yield SkippedDependency(
+                    name=req.requirement_line.line,
+                    skip_reason="could not deduce package/specifier pair from requirement, "
+                    "please specify them with #egg=your_package_name==your_package_version",
```
woodruffw, line 248: I think we need to correct this advice: `#egg=name==version` isn't `pip`-supported syntax. AFAICT, there's no way to pre-declare the package version for a PEP-508-style URL requirement -- the distribution always has to be fetched.

xref: https://github.com/pypa/pip/pull/11617

pip_audit/_dependency_source/requirement.py
woodruffw: Thanks for highlighting this! I agree that this isn't ideal -- `--require-hashes` should provide *more* consistency and accuracy during dependency audits, not less. 

> I think there's a future discussion to be had about whether it makes sense to keep this "preresolved" optimisation since pip install could be more efficient than our previous homegrown resolver. The new method won't have the problem we used to have where we potentially install a sdist for each candidate just to figure out its dependencies.

Yeah, let's open up a discussion issue for this. I'm not opposed to removing this special case if the performance hit isn't as significant as the regression we had with `2.4.15`.

woodruffw (approved):
LGTM, thanks @tetsuo-cpp! It'll be great to send a release out with this.

PR #538, opened 2023-03-10
requirement: Support process substitution inputs

### LINKED ISSUE #537: Regression: handling process substitution inputs
This regressed with #523:

```console
$ pip-audit -r <(echo 'something')

ERROR:pip_audit._virtual_env:internal pip failure: ERROR: Could not open requirements file: [Errno 9] Bad file descriptor: '/dev/fd/63'
```

`<(...)` is a bash-ish for process substitution: it turns the output of the subshell into a `/dev/fd/<n>` path, which the process can then read from.

My guess is that this regressed because of how we're plumbing inputs into the `pip` subprocess: we probably now pass the filenames in directly during `VirtualEnv` construction. Unfortunately that doesn't work for `/dev/fd/<n>` inputs, since they're unique to the parent process.

To fix this, we probably need to buffer or otherwise copy our requirement inputs.

tetsuo-cpp:
Closes #537

### DIFF
--- a/pip_audit/_dependency_source/requirement.py
+++ b/pip_audit/_dependency_source/requirement.py
@@ -97,16 +97,16 @@ class RequirementSource(DependencySource):
             # In order to get around this, we're going to copy each input into a
             # a corresponding temporary file and then pass that set of files
             # into `pip`.
-            tmp_files: list[IO[str]] = [
-                stack.enter_context(NamedTemporaryFile(mode="w")) for _ in self._filenames
-            ]
+            tmp_files = []
 
             # For each input file, copy it to one of our temporary files.
-            # Ensure we flush and fsync so our writes are visible to `pip`.
-            for filename, tmp_file in zip(self._filenames, tmp_files):
+            # Ensure we flush so our writes are visible to `pip`.
+            for filename in self._filenames:
+                tmp_file = stack.enter_context(NamedTemporaryFile(mode="w"))
                 with filename.open("r") as f:
                     shutil.copyfileobj(f, tmp_file)
                     tmp_file.flush()
+                tmp_files.append(tmp_file)
 
             # Now pass the list of temporary filenames into the rest of our
             # logic.
--- a/pip_audit/_dependency_source/requirement.py
+++ b/pip_audit/_dependency_source/requirement.py
@@ -84,10 +84,40 @@ class RequirementSource(DependencySource):
         Raises a `RequirementSourceError` on any errors.
         """
 
+        with ExitStack() as stack:
+            # We need to handle process substitution inputs so we can invoke
+            # `pip-audit` like so:
+            #
+            #   pip-audit -r <(echo 'something')
+            #
+            # Since `/dev/fd/<n>` inputs are unique to the parent process, we
+            # can't pass these file names to `pip` and expect `pip` to able to
+            # read them.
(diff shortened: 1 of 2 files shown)


tetsuo-cpp:
I didn't add anything to the CHANGELOG since this is basically fixing an unreleased regression.

pip_audit/_dependency_source/requirement.py
```diff
@@ -84,10 +84,40 @@ def collect(self) -> Iterator[Dependency]:
...
+                stack.enter_context(NamedTemporaryFile(mode="w")) for _ in self._filenames
+            ]
+
+            # For each input file, copy it to one of our temporary files.
+            # Ensure we flush and fsync so our writes are visible to `pip`.
+            for filename, tmp_file in zip(self._filenames, tmp_files):
+                with filename.open("r") as f:
+                    shutil.copyfileobj(f, tmp_file)
+                    tmp_file.flush()
+                    os.fsync(tmp_file)
```
woodruffw, line 110: Do we need both of these? I would expect one or the other to be sufficient, but POSIX I/O never fails to surprise...

pip_audit/_dependency_source/requirement.py
```diff
@@ -84,10 +84,40 @@ def collect(self) -> Iterator[Dependency]:
...
+            # In order to get around this, we're going to copy each input into a
+            # a corresponding temporary file and then pass that set of files
+            # into `pip`.
+            tmp_files: list[IO[str]] = [
+                stack.enter_context(NamedTemporaryFile(mode="w")) for _ in self._filenames
+            ]
+
+            # For each input file, copy it to one of our temporary files.
+            # Ensure we flush and fsync so our writes are visible to `pip`.
+            for filename, tmp_file in zip(self._filenames, tmp_files):
```
woodruffw, line 106: Minor thing, but rather than two loops and a `zip`, maybe we can just do a single loop here? Something like:

```python
tmp_files = []
for filename in self._filenames:
    tmp_file = stack.enter_context(...)
    with filename.open("r") as io:
        ...
        
# use tmp_files here
```

pip_audit/_dependency_source/requirement.py
tetsuo-cpp: Oh good point. I think those are both required for the write to be durable, but we don't care about the temporary file persisting if we power off anyway. I think I can get away with just `flush`.

pip_audit/_dependency_source/requirement.py
tetsuo-cpp: Yep, that works.

pip_audit/_dependency_source/requirement.py
tetsuo-cpp: [3bb322b](https://github.com/pypa/pip-audit/pull/538/commits/3bb322bc905f6fcc8f53ca3787abfd7a25011b45)

pip_audit/_dependency_source/requirement.py
tetsuo-cpp: [b14e1b5](https://github.com/pypa/pip-audit/pull/538/commits/b14e1b55942b606db97efc60ae2dc31f8cb63bf0)

woodruffw (approved):
LGTM!

PR #549, opened 2023-03-16
workflows/release: use OIDC exchange

woodruffw:
Now that we've merged this upstream 🙂

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -5,17 +5,17 @@ on:
 
 name: release
 
-permissions:
-  # Used to sign the release's artifacts with sigstore-python.
-  id-token: write
-
-  # Used to attach signing artifacts to the published release.
-  contents: write
-
 jobs:
   pypi:
     name: upload release to PyPI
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -36,7 +38,6 @@ jobs:
       uses: pypa/gh-action-pypi-publish@release/v1
       with:
         user: __token__
-        password: ${{ secrets.PYPI_TOKEN }}
```
di: We can remove all of these lines, right?

woodruffw: Yep! I missed that last one; just removed.

woodruffw:
I've also deleted the `PYPI_TOKEN` secret. @di I think you'll have to remove the corresponding token on the PyPI side, since it's not tied to my account (AFAICT).

di:
@woodruffw I'm not seeing it on my account either.

woodruffw:
Yeah, my mistake -- it was named poorly on my side. It's removed now!

PR #565, opened 2023-03-22
_dependency_source, _cli, test: Do not override `pip.conf` unless explicitly specified via flags

### LINKED ISSUE #563: Private repo dependencies no longer being skipped
## Bug description

Private repo dependencies are no longer being skipped if they cannot be resolved.
For example, in the v2.4.14, the audit would succeed and then show that the packages were skipped:
image

But now in v2.5.x, the audit fails and shows the error:
image

## Reproduction steps

- Ensure `pip-audit` v2.5.x is installed
- Create a `requirements.txt` file with a dependency that is installed from a private repo
- Run `pip-audit -r requirements.txt`

## Expected behavior

I'd expect the same behaviour as v2.4.14 - whereby the private dependencies would simply be skipped - unless the `-S` flag was specified.

## Platform information

* OS name and version:
* `pip-audit` version (`pip-audit -V`): 2.5.2
* Python version (`python -V` or `python3 -V`): 3.11.2
* `pip` version (`pip -V` or `pip3 -V`): 23.0.1

philblckwd:
Refactor index-url and extra-index-url options to not override config defined in 'pip.conf' unless explicitly specified via flags.

Closes #563

### DIFF
--- a/pip_audit/_cli.py
+++ b/pip_audit/_cli.py
@@ -284,7 +284,7 @@ def _parser() -> argparse.ArgumentParser:  # pragma: no cover
         "--index-url",
         type=str,
         help="base URL of the Python Package Index; this should point to a repository compliant "
-        "with PEP 503 (the simple repository API)",
+        "with PEP 503 (the simple repository API); this will be resolved by pip if not specified",
     )
     parser.add_argument(
         "--extra-index-url",
--- a/pip_audit/_cli.py
+++ b/pip_audit/_cli.py
@@ -16,7 +16,6 @@ from typing import IO, Iterator, NoReturn, cast
 from pip_audit import __version__
 from pip_audit._audit import AuditOptions, Auditor
 from pip_audit._dependency_source import (
-    PYPI_URL,
     DependencySource,
     DependencySourceError,
     PipSource,
@@ -286,7 +285,6 @@ def _parser() -> argparse.ArgumentParser:  # pragma: no cover
         type=str,
         help="base URL of the Python Package Index; this should point to a repository compliant "
         "with PEP 503 (the simple repository API)",
-        default=PYPI_URL,
     )
     parser.add_argument(
         "--extra-index-url",
@@ -347,12 +345,14 @@ def _parse_args(parser: argparse.ArgumentParser) -> argparse.Namespace:  # pragm
 
 
 def _dep_source_from_project_path(
-    project_path: Path, index_urls: list[str], state: AuditState
+    project_path: Path, index_url: str, extra_index_urls: list[str], state: AuditState
 ) -> DependencySource:  # pragma: no cover
     # Check for a `pyproject.toml`
     pyproject_path = project_path / "pyproject.toml"
     if pyproject_path.is_file():
-        return PyProjectSource(pyproject_path, index_urls=index_urls, state=state)
+        return PyProjectSource(
+            pyproject_path, index_url=index_url, extra_index_urls=extra_index_urls, state=state
+        )
 
     # TODO: Checks for setup.py and other project files will go here.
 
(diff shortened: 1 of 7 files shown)


pip_audit/_cli.py
```diff
@@ -286,7 +285,6 @@ def _parser() -> argparse.ArgumentParser:  # pragma: no cover
         type=str,
         help="base URL of the Python Package Index; this should point to a repository compliant "
         "with PEP 503 (the simple repository API)",
-        default=PYPI_URL,
```
woodruffw, line 289: I think we want to keep this `default` in the CLI -- behaviorally it should remain the default, and having it appear as such in the `--help` render makes for a nicer UX 🙂

pip_audit/_dependency_source/pyproject.py
```diff
@@ -33,14 +28,20 @@ class PyProjectSource(DependencySource):
     """
 
     def __init__(
-        self, filename: Path, index_urls: list[str] = [PYPI_URL], state: AuditState = AuditState()
+        self,
+        filename: Path,
+        index_url: str = "",
```
woodruffw, line 33: Similar typestate nit here: maybe drop the default `""` here (or make it `PYPI_URL`), since we always expect a valid URL?

pip_audit/_dependency_source/requirement.py
```diff
@@ -47,7 +46,8 @@ def __init__(
         require_hashes: bool = False,
         no_deps: bool = False,
         skip_editable: bool = False,
-        index_urls: list[str] = [PYPI_URL],
+        index_url: str = "",
```
woodruffw, line 49: Same as https://github.com/pypa/pip-audit/pull/565/files#r1145039056

woodruffw:
```suggestion
        index_url: str | None = None,
```

woodruffw:
Thanks a ton for opening this, @philblckwd! 

This looks great to me structurally; just a few small nitpicks 🙂

woodruffw:
(This also needs a `CHANGELOG.md entry. Feel free to do it yourself if you feel like it, otherwise I'll do it as a fixup after merge!)

pip_audit/_cli.py
philblckwd: So leaving this as the default means that it will always be explicitly passed to the `pip install` CLI, which results in the `index-url` specified in the `pip.conf` being ignored. I understand that it makes for a more friendly UX to specify as such, but would it be reasonable to convey that the url will be resolved by `pip` if not passed by the user explicitly?

If this is a deal breaker though, then I'll have to dig deeper into `pip` to determine how it is performing the `index-url` resolution.

pip_audit/_dependency_source/pyproject.py
tetsuo-cpp: Yep, I'd prefer this too.

tetsuo-cpp:
I'm also happy with where this is at. Once the help text is amended and those type state related comments are addressed, this should be good to go.

Thanks again for your work. We really appreciate it.

tetsuo-cpp:
@philblckwd One more thing. We have a check in CI to ensure that the help text that we display in the README is up to date. So once you've amended the help text for `--index-url`, you'll need to update the README.

philblckwd:
Per feedback:

- I updated arguments to use optional types
- I updated the help text, as well as the readme, to state the `index-url` will be resolved by pip if not specified
- I added an entry to the changelog
- I noticed unrelated lint and check-readme errors, so resolved those too

tetsuo-cpp (approved):
LGTM!

tetsuo-cpp:
I'll merge this shortly after I've done a bit of testing locally.

tetsuo-cpp:
Thanks @philblckwd!

PR #567, opened 2023-03-23
_subprocess: Fix `pip install` log window not showing

tetsuo-cpp:
Related to https://github.com/pypa/pip-audit/pull/536. I noticed this when debugging something else. For some reason, when we read `stderr` in the polling loop, it causes us to read nothing for `stdout`. There's no reason that we need to read `stderr` in the loop since it's not getting dumped into the log window so I just moved this code out of the loop.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,10 @@ All versions prior to 0.0.9 are untracked.
 * Fixed a crash on Windows caused by multiple open file handles to
   input requirements ([#551](https://github.com/pypa/pip-audit/pull/551))
 
+* Fixed an issue where the log window that we use to display `pip-audit`'s
+  dependency resolution progress was not showing anything
+  ([#567](https://github.com/pypa/pip-audit/pull/567))
+
 ## [2.5.0]
 
 ### Changed
--- a/pip_audit/_subprocess.py
+++ b/pip_audit/_subprocess.py
@@ -57,6 +57,9 @@ def run(args: Sequence[str], *, log_stdout: bool = False, state: AuditState = Au
             f"Running {pretty_args}", stdout.decode(errors="replace") if log_stdout else None
         )
 
+    # NOTE(alex): For reasons I'm unsure about, reading the stderr stream in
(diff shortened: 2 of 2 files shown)


CHANGELOG.md
```diff
@@ -51,6 +51,10 @@ All versions prior to 0.0.9 are untracked.
 * Fixed a crash on Windows caused by multiple open file handles to
   input requirements ([#551](https://github.com/pypa/pip-audit/pull/551))
 
+* Fixed an issue where the log window that we use to display `pip-audit`'s
+  dependency resolution progress was not showing anything
+  ([#567](https://github.com/pypa/pip-audit/pull/567))
```
woodruffw, line 56: Needs relocating under `[Unreleased]` 🙂

pip_audit/_subprocess.py
```diff
@@ -54,11 +53,14 @@ def run(args: Sequence[str], *, log_stdout: bool = False, state: AuditState = Au
...
         stdout += process.stdout.read(4096)  # type: ignore
-        stderr += process.stderr.read(4096)  # type: ignore
         state.update_state(
             f"Running {pretty_args}", stdout.decode(errors="replace") if log_stdout else None
         )
 
+    # NOTE(alex): For reasons I'm unsure about, reading the stderr stream in
+    # real time seems to interfere with stdout and cause us to read nothing.
+    # Let's wait until the process is terminated before reading stderr.
+    stderr = process.stderr.read()  # type: ignore
```
woodruffw, line 63: Hmm, I'm a little worried this might cause deadlocks in some (unlikely) conditions: I can imagine a case where `pip` writes to `stderr` before `stdout` and blocks waiting for the parent's corresponding read, meaning that `process.poll()` never sees a process termination (and `stdout.read(bufsize)` always results in an empty read).

Specifically, that should only happen if `pip` writes enough to `stderr` to saturate the underlying pipe buffer. But that _might_ happen, so it's probably worth root causing what's going on here instead of deferring the `stderr` read.

tetsuo-cpp:
This code is going to be reworked as part of #574 so let's see if that fixes it.

PR #596, opened 2023-04-18
OS matrix in CI

### LINKED ISSUE #553: Run tests for Windows in GitHub Actions
We recently had a platform specific regression on Windows (https://github.com/pypa/pip-audit/issues/548). Our CI runs on Linux and most of our development happens on MacOS, so these types of bugs could happen again the future. It might be worth looking into running our testing on Windows.

jleightcap:
Resolves #553

### DIFF
--- a/test/test_cache.py
+++ b/test/test_cache.py
@@ -1,3 +1,4 @@
+import os
 from pathlib import Path
 
 import pretend  # type: ignore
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -6,25 +6,30 @@ on:
...
+          - { py: "3.11", os: "ubuntu-latest" }
+          # NOTE: We only test Windows and macOS on the latest Python;
+          # these primarily exist to ensure that we don't accidentally
+          # introduce Linux-isms into the development tooling.
+          - { py: "3.11", os: "windows-latest" }
+          - { py: "3.11", os: "macos-latest" }
+    runs-on: ${{ matrix.conf.os }}
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@83b7061638ee4956cf7545a6f7efe594e5ad0247 # v3.5.1
```
woodruffw, line 28: IMO no need to pin to hashes here -- these are official actions, which makes them as close to "blessed" as possible.

woodruffw: Yeah, `sigstore-python` uses a stricter policy. Let's drop down to a symbolic version here.

woodruffw, line 28:
```suggestion
      - uses: actions/checkout@v3
```

test/test_cache.py
```diff
@@ -9,15 +10,15 @@
 
 def test_get_cache_dir(monkeypatch):
     # When we supply a cache directory, always use that
-    cache_dir = _get_cache_dir(Path("/tmp/foo/cache_dir"))
-    assert str(cache_dir) == "/tmp/foo/cache_dir"
+    cache_dir = _get_cache_dir(os.path.join("tmp", "foo", "cache_dir"))
+    assert str(cache_dir) == os.path.join("tmp", "foo", "cache_dir")
```
woodruffw, line 14: What's happening here?

woodruffw: Oh, is this because of the Windows failure? We should probably use pytest's built-in tempdir functionality here 🙂

woodruffw: Ref: https://docs.pytest.org/en/7.1.x/how-to/tmp_path.html

PR #605, opened 2023-05-03
pyproject: add, pin requests

woodruffw:
Looks like we were implicitly depending on this before, despite explicitly importing it, which is bad!

This adds a `2.29.*` constraint on our `requests` dep, which should block off a regression in `CacheControl` due to underlying changes in `urllib3` for the time being.

xrefs:

* https://github.com/psf/requests/issues/6437
* https://github.com/ionrock/cachecontrol/issues/292

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -35,6 +35,7 @@ dependencies = [
     "packaging>=23.0.0",                     # https://github.com/pypa/pip-audit/issues/464
     "pip-api>=0.0.28",
     "pip-requirements-parser>=32.0.0",
+    "requests==2.29.*",
     "rich>=12.4",
     "toml>=0.10",
 ]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -35,7 +35,8 @@ dependencies = [
     "packaging>=23.0.0",                     # https://github.com/pypa/pip-audit/issues/464
     "pip-api>=0.0.28",
     "pip-requirements-parser>=32.0.0",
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -35,6 +35,9 @@ dependencies = [
     "packaging>=23.0.0",                     # https://github.com/pypa/pip-audit/issues/464
     "pip-api>=0.0.28",
     "pip-requirements-parser>=32.0.0",
+    # NOTE(ww): requests 2.30.0 and higher are incompatible with CacheControl;
+    # See: https://github.com/psf/requests/issues/6437
+    "requests==2.29.*",
```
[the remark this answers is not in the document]
woodruffw: No good reason, can relax.

di (approved):

pyproject.toml
```diff
@@ -35,6 +35,9 @@ dependencies = [
     "packaging>=23.0.0",                     # https://github.com/pypa/pip-audit/issues/464
     "pip-api>=0.0.28",
     "pip-requirements-parser>=32.0.0",
+    # NOTE(ww): requests 2.30.0 and higher are incompatible with CacheControl;
```
sethmlarson, line 38: This comment isn't quite correct, it's urllib3 v2.0 that isn't compatible with CacheControl: https://github.com/psf/requests/issues/6437#issuecomment-1533357246

woodruffw: Thanks! Will fix.

Issue #575, opened 2023-03-23
Bump our CycloneDX dependency


woodruffw:
Looks like there's a 4.0 series of CycloneDX for Python now: <https://github.com/CycloneDX/cyclonedx-python-lib/releases/tag/v4.0.0>

We should bump our dependency and rewrite our API usage to update to it.

woodruffw:
#625

PR #625, opened 2023-05-30
updated cyclonedx dependency

ThunderKey:
I've had a quick look at #575 and have seen that it is quite straight  forward. Therefore, I've had a go at it.

To check that the same result is generated I've updated a test to check that the JSON is the same. The only change in the generated JSON was that the vulnerabilities got sorted and in the `dependencies` the value `'dependsOn': [],` is missing.

Since it is not the aim of the test to test the exact output I did not commit it. If you are interested in the test I changed I can submit it as well. https://github.com/pypa/pip-audit/blob/f7a22cf193569aaa6ad2c00874ebf07fda65ddd5/test/format/test_cyclonedx.py#L21

Additionally, I don't think that the `_PipAuditResultParser` is necessary and it could be a function, because the `BaseParser` does not really add functionality. If  you feel the same way I can refactor it. https://github.com/pypa/pip-audit/blob/f7a22cf193569aaa6ad2c00874ebf07fda65ddd5/pip_audit/_format/cyclonedx.py#L24

### DIFF
--- a/pip_audit/_format/cyclonedx.py
+++ b/pip_audit/_format/cyclonedx.py
@@ -24,6 +24,7 @@ logger = logging.getLogger(__name__)
 class _PipAuditResultParser(BaseParser):
     def __init__(self, result: dict[service.Dependency, list[service.VulnerabilityResult]]):
         super().__init__()
+        self.vulnerabilities = []
 
         for dep, vulns in result.items():
             # TODO(alex): Is there anything interesting we can do with skipped dependencies in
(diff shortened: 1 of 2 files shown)


woodruffw:
Thanks a ton @ThunderKey!

I agree that testing the exact output is not important here: our CycloneDX compatibility guarantee is only that we comply with the schema, and nothing beyond that. Given that you haven't bumped the schema in this PR, any breaking changes here would be a bug on CycloneDX's side (if they are bugs).

> Additionally, I don't think that the _PipAuditResultParser is necessary and it could be a function, because the BaseParser does not really add functionality. If you feel the same way I can refactor it.

I'm not opposed to that -- IIRC we did it that way because it was recommended by the library's documentation, but I agree that it isn't necessary.

That being said, I'd prefer to keep this PR hermetic. If you feel like doing that, let's do it in a separate PR 🙂

woodruffw (approved):

woodruffw:
LGTM, thank you!

PR #540, opened 2023-03-10
requirement, test: Remove preresolved dependency optimization

tetsuo-cpp:
Closes #539.

Closes #564.

PR #626, opened 2023-05-31
Reintroduce dependency optimization

### LINKED ISSUE #610: Reintroduce the preresolved dependency optimization
We removed this in #540 for correctness reasons.

However, doing so seems to have caused some bad performance regressions on `brew-pip-audit`'s automation: audits that were previously only bound by vulnerability API lookup time are now much slower, as they call into `pip` and do dependency machinery work.

We should consider reintroducing a variant of these, potentially stuffed behind a scary flag.

CC @tetsuo-cpp for thoughts.

CC @alex for visiblity.

trottomv:
Fix #610

### DIFF
--- a/pip_audit/_cli.py
+++ b/pip_audit/_cli.py
@@ -327,6 +327,13 @@ def _parser() -> argparse.ArgumentParser:  # pragma: no cover
             "this option can be used multiple times"
         ),
     )
+    parser.add_argument(
+        "--disable-pip",
+        action="store_true",
+        help="don't use `pip` for dependency resolution; "
+        "this can only be used with hashed requirements files or if the `--no-deps` flag has been "
+        "provided",
+    )
     return parser
 
 
@@ -380,10 +387,15 @@ def audit() -> None:  # pragma: no cover
             parser.error("The --extra-index-url flag can only be used with --requirement (-r)")
         elif args.no_deps:
             parser.error("The --no-deps flag can only be used with --requirement (-r)")
+        elif args.disable_pip:
+            parser.error("The --disable-pip flag can only be used with --requirement (-r)")
 
     # Nudge users to consider alternate workflows.
     if args.require_hashes and args.no_deps:
         logger.warning("The --no-deps flag is redundant when used with --require-hashes")
+
+    if args.no_deps and args.disable_pip:
+        logger.warning("The --no-deps flag is redundant without --disable-pip")
 
     if args.require_hashes and isinstance(service, OsvService):
         logger.warning(
@@ -414,6 +426,7 @@ def audit() -> None:  # pragma: no cover
                 req_files,
                 require_hashes=args.require_hashes,
                 no_deps=args.no_deps,
+                disable_pip=args.disable_pip,
                 skip_editable=args.skip_editable,
                 index_url=args.index_url,
                 extra_index_urls=args.extra_index_urls,
--- a/pip_audit/_cli.py
+++ b/pip_audit/_cli.py
@@ -395,7 +395,7 @@ def audit() -> None:  # pragma: no cover
         logger.warning("The --no-deps flag is redundant when used with --require-hashes")
 
     if args.no_deps and args.disable_pip:
-        logger.warning("The --no-deps flag is redundant without --disable-pip")
+        logger.warning("The --no-deps flag is redundant when used with --disable-pip")
 
     if args.require_hashes and isinstance(service, OsvService):
         logger.warning(
--- a/pip_audit/_dependency_source/requirement.py
+++ b/pip_audit/_dependency_source/requirement.py
@@ -46,6 +46,7 @@ class RequirementSource(DependencySource):
         *,
         require_hashes: bool = False,
         no_deps: bool = False,
+        disable_pip: bool = False,
         skip_editable: bool = False,
         index_url: str | None = None,
         extra_index_urls: list[str] = [],
@@ -59,9 +60,13 @@ class RequirementSource(DependencySource):
         `require_hashes` controls the hash policy: if `True`, dependency collection
         will fail unless all requirements include hashes.
 
-        `no_deps` controls the dependency resolution policy: if `True`,
+        `disable_pip` controls the dependency resolution policy: if `True`,
         dependency resolution is not performed and the inputs are checked
         and treated as "frozen".
+
+        `no_deps` controls whether dependency resolution can be disabled even without
+        hashed requirements (which implies a fully resolved requirements file): if `True`,
+        `disable_pip` is allowed without a hashed requirements file.
 
         `skip_editable` controls whether requirements marked as "editable" are skipped.
         By default, editable requirements are not skipped.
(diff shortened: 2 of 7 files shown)


pip_audit/_cli.py
```diff
@@ -327,6 +327,12 @@ def _parser() -> argparse.ArgumentParser:  # pragma: no cover
             "this option can be used multiple times"
         ),
     )
+    parser.add_argument(
+        "--with-pip-compile",
+        action="store_true",
+        help="don't perform dependency installation in a temporary virtual env"
+        " if requirements files is builded with pip compile; requires --requirement argument",
+    )
```
woodruffw, line 335: This implies that hashed requirements files always come from `pip-compile`, which isn't true. 

Rather than adding a new flag here, let's revert the behavior that was removed with #540: the pre-existing `--no-deps` and `--require-hashes` flags should cause us to go down an optimized path.

pip_audit/_dependency_source/requirement.py
```diff
@@ -44,6 +53,7 @@ def __init__(
         index_url: str | None = None,
         extra_index_urls: list[str] = [],
         state: AuditState = AuditState(),
+        with_pip_compile: bool = False,
```
woodruffw, line 56: Same comment as above: I don't think we want to imply that this is unique to `pip-compile`, since it wasn't before.

pip_audit/_dependency_source/requirement.py
```diff
@@ -122,7 +133,19 @@ def collect(self) -> Iterator[Dependency]:
...
+        # Parse requirements files without install in temporary virtual env.
+        parsed_packages = []
+        for filename in filenames:
+            requirements_file = RequirementsFile.from_file(filename)
+            requirements_dict = requirements_file.to_dict()["requirements"]
+            parsed_packages += [
+                (r["name"], r["specifier"][0].split("==")[1]) for r in requirements_dict
+            ]
+        for name, version in parsed_packages:
+            yield ResolvedDependency(name=name, version=Version(version))
```
woodruffw, line 146: This needs to be hardened against URL-style requirements, parser errors, etc.

My suggestion would be to look at what was removed in #540, and reuse that to the greatest extent possible: it had a lot of "scar tissue" from different bugs reported over time.

woodruffw (changes requested):
Thanks again for opening this! 

This is a good start, but it's going to need a few changes before I think we can consider merging this.

In particular, my recommendation is to go back and look at the optimization's state at the time it was removed in #540 -- the re-addition of the optimization here should closely mirror that code, rather than being an independent reimplementation. That will save us a lot of heartburn and trouble down the road.

trottomv:
@woodruffw I've removed the "with_pip_compile" option and restored the preresolved dependencies collection as in #540

woodruffw:
Cool, thank you! I'll have time to review again in the coming days.

pauloxnet:
It seems ok to me. Maybe you can update the Changelog file.

woodruffw:
Lint failure:

```
would reformat pip_audit/_dependency_source/interface.py
would reformat pip_audit/_dependency_source/requirement.py
would reformat test/dependency_source/test_requirement.py

Oh no! 💥 💔 💥
3 files would be reformatted, 48 files would be left u
```

You can use `make reformat` and `make lint` locally to catch these.

pauloxnet (approved):

pauloxnet:
LGTM

woodruffw (approved):
LGTM; I'll do some end-CLI testing this evening.

I'd like @tetsuo-cpp to also give this a review, since he wrote the original fast-path here.

tetsuo-cpp:
Sorry for the delay @trottomv! I'll get to this tonight.

pip_audit/_dependency_source/requirement.py
```diff
@@ -123,11 +129,30 @@ def collect(self) -> Iterator[Dependency]:
...
+                )
+
+            # If one or more requirements have a hash, this implies `--require-hashes`.
+            require_hashes = require_hashes or any(req.hash_options for req in rf.requirements)
+            reqs.extend(rf.requirements)
+
+        # If the user has supplied `--no-deps` or there are hashed requirements, we should assume
+        # that we have a fully resolved set of dependencies and we should waste time by invoking
+        # `pip`.
+        if self._no_deps or require_hashes:
```
tetsuo-cpp, line 151: @woodruffw I thought the plan was that we were going to give it some kind of scary name instead of having it activate as soon as we supply a hashed requirement since the fast path is less comprehensive than doing a real `pip` installation.

tetsuo-cpp (approved):
Code looks good!

@woodruffw Regarding my comment about the flag, I don't feel too strongly about it either way, it just wasn't what I was expecting. If that seems good to you, feel free to merge.

pip_audit/_dependency_source/requirement.py
trottomv: Hi @tetsuo-cpp thanks a lot for your review.
I agree with you. In my humble opinion, invoking "preresolved dependency" based solely on the requirement presenting the hashes is a bit weak. My original idea was to add a new flag, and I explained it better in this comment on the issue. https://github.com/pypa/pip-audit/issues/610#issuecomment-1558811701

woodruffw: @tetsuo-cpp You're right, I forgot about that -- IMO `--disable-pip` or similar would be appropriate.

The semantics there would be:

* `--disable-pip` will cause an error unless we're in a requirements input mode
* `--disable-pip` will cause an error unless the requirements are fully hashed *or* the user also passes `--no-deps`

Does that sound reasonable to you?

@trottomv I'd prefer to merge this as a single feature, so we'll probably want to settle the new flag before merging 🙂

trottomv: No problem for me, if I can contribute, I'm happy to do so, time permitting

tetsuo-cpp: @woodruffw Yep, those flag names seem fine to me! Sounds good.

@trottomv Sorry for the churn! Are you able to make those changes? If not, I can find some time and make some pushes to this branch.

tetsuo-cpp: @woodruffw Yep! Those flag names sound good to me.

@trottomv Sorry for the churn! I'm planning to push those change to your branch this afternoon and we can look at getting this released.

tetsuo-cpp:
@woodruffw, I've made those changes now. Could you take a quick look at this?

pip_audit/_cli.py
```diff
@@ -380,11 +387,16 @@ def audit() -> None:  # pragma: no cover
...
             parser.error("The --no-deps flag can only be used with --requirement (-r)")
+        elif args.disable_pip:
+            parser.error("The --disable-pip flag can only be used with --requirement (-r)")
 
     # Nudge users to consider alternate workflows.
     if args.require_hashes and args.no_deps:
         logger.warning("The --no-deps flag is redundant when used with --require-hashes")
 
+    if args.no_deps and args.disable_pip:
+        logger.warning("The --no-deps flag is redundant without --disable-pip")
```
woodruffw, line 398: Minor typo? `without -> with`:

```suggestion
        logger.warning("The --no-deps flag is redundant when used with --disable-pip")
```

woodruffw (approved):
LGTM!

woodruffw:
Thanks for all of the hard work here @trottomv! Merging.

PR #647, opened 2023-07-06
Fix tempfiles on windows, fixes #646

### LINKED ISSUE #646: Tempfiles are not working on Windows
## Bug description
On Windows, when `pip-audit` creates the virtual environment a permission error occurs. This is caused under the hood by a [known bug in Python](https://stackoverflow.com/questions/23212435/permission-denied-to-write-to-my-temporary-file) itself. However, there is a way to work around it (see the end of this issue report).

During the pip-audit process a virtual environment is created in a temporary folder. A temporary file created in this temp folder loses its permissions on Windows when it is closed (this doesn't happen in Unix). The pip install process opens and closes one or two temporary files several times during the installation process, and thus, permissions on the temp file are lost and Windows throws an error.

## Reproduction steps
On windows for any python package:
```
pip-audit .
``` 

## Expected behavior
Create virtual environment for auditing.

## Screenshots and logs
The error message
```
ERROR:pip_audit._virtual_env:internal pip failure: ERROR: Could not open requirements file: [Errno 13] Permission denied: 'C:\\Users\\Manrho\\AppData\\Local\\Temp\\82\\tmpnmkebs43'

ERROR:pip_audit._cli: Failed to install packages: ['C:\\Users\\Manrho\\AppData\\Local\\Temp\\82\\tmp855pujcg\\Scripts''python.exe', ' -m', ' pip', ' install', ' --dry-run', '--report', 'C:\\Users\\Manrho\\AppData\\Local\\Temp\\82\\tmpm_c5v3cp', '-r', 'C:\\Users\\Manrho\\AppData\\Local\\Temp\\82\\tmpnmkebs43']
```

## Platform information
* OS name and version: Windows Server 2019
* `pip-audit` version (`pip-audit -V`): 2.6.0
* Python version (`python -V` or `python3 -V`): 3.9.17
* `pip` version (`pip -V` or `pip3 -V`): 23.1.2

## Additional context
An easy fix is to replace `NamedTemporaryFile()` to `NamedTemporaryFile(delete=False)` in the files `_virtual_env.py` and `pip_audit/_dependency_source'. However, this defeats the purpose of using NamedTemporaryFile as manual cleanup is needed.
The way to fix this is to avoid using the `NamedTemporaryFile()` context manager altogether.

marickmanrho:
This PR fixes pip-audit on Windows machines by providing a work around for `NamedTempFile`. This workaround is implemented in the `_util.py` file as `CustomNamedTempFile`.

The `CustomNamedTempFile` creates a temporary file as expected but allows the file to be open and closed multiple times. `CustomNamedTempFile` also works as a contextmanager and closes+removes the file when exiting the context.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,11 @@ All versions prior to 0.0.9 are untracked.
 
 ## [Unreleased]
 
+### Fixed
+
+* Fixed a crash on Windows caused by `pip-audit`'s use of temporary files
+  ([#647](https://github.com/pypa/pip-audit/pull/647))
+
 ## [2.6.0]
 
 ### Added


pip_audit/_util.py
```diff
@@ -3,11 +3,36 @@
...
 
 import sys
+import tempfile
+import os
 from typing import NoReturn  # pragma: no cover
 
 from packaging.version import Version
 
 
+class CustomNamedTemporaryFile:
```
woodruffw, line 13: We control the `pip_audit` namespace, so I don't think we need the `Custom` namespace:

```suggestion
class NamedTemporaryFile:
```

and then just `from pip_audit._util import NamedTemporaryFile` wherever we need it 🙂

pip_audit/_util.py
```diff
@@ -3,11 +3,36 @@
...
+        # Ensure the file is created
+        if not os.path.exists(file_name):
+            open(file_name, 'x').close()
+        # Open the file in the given mode
+        self._tempFile = open(file_name, self._mode)
+        return self._tempFile
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        self._tempFile.close()
+        os.unlink(self._tempFile.name)
```
woodruffw, line 34: A few thoughts:

* Rather than recreating most of the internals of `tempfile.NamedTemporaryFile`, can we re-use `tempfile.NamedTemporaryFile(delete=False)` for the file object here?
* Even if we don't use the "real" `NamedTemporaryFile` here, we should try to keep divergences from the "normal" tempfile behavior to a minimum. In particular, `os.urandom(24).hex()` looks like a reimplementation of `mkstemp`.

woodruffw (commented):
Sorry for the delay here; a couple of thoughts!

pip_audit/_dependency_source/pyproject.py
```diff
@@ -76,21 +76,26 @@ def collect(self) -> Iterator[Dependency]:
...
-                try:
-                    ve.create(ve_dir)
-                except VirtualEnvError as exc:
-                    raise PyProjectSourceError(str(exc)) from exc
-
-                # Now query the installed packages.
-                for name, version in ve.installed_packages:
-                    yield ResolvedDependency(name=name, version=version)
+            with TemporaryDirectory() as ve_dir:
+                with NamedTemporaryFile(dir=ve_dir, delete=False) as req_file:
```
woodruffw, line 80: Nitpick: use a chained `with` statement to make this terser:

```suggestion
            with TemporaryDirectory() as td, NamedTemporaryFile(dir=td, delete=False) as req_file:
```

pip_audit/_virtual_env.py
```diff
@@ -110,38 +110,45 @@ def post_setup(self, context: SimpleNamespace) -> None:
...
-
-            # Convert into a series of name, version pairs
-            self._packages = []
-            for package in package_list:
-                package_metadata = package["metadata"]
-                self._packages.append(
-                    (package_metadata["name"], Version(package_metadata["version"]))
-                )
+        with TemporaryDirectory() as ve_dir:
+            with NamedTemporaryFile(dir=ve_dir, delete=False) as tmp:
```
woodruffw, line 114: Same nitpick as above 🙂

woodruffw (approved):
LGTM, thank you @marickmanrho!

woodruffw:
This is included in 2.6.1, which is in the process of being released now. Thanks again!

PR #653, opened 2023-07-26
mvp for getting scores and severity of vulnerabilities

matteoannotell:
Purpose: we want to retrieve a measure of severity of the vulnerabilities detected, in order to suppress too noisy alerts.
To do so, we introduce an API call to 
```
base_url = "https://services.nvd.nist.gov/rest/json/cve/1.0/"
url = f"{base_url}/{vulnerability_id}"
````
for each CVE vulnerability detected, either via pypi or via osv. 
When the CVE is not present or not found in the database, we return None.

The layout of the tabular output has been changed accordingly to include two new columns, `score` and `severity`, which are respectively 
```
first_item["impact"]["baseMetricV3"]["cvssV3"]["baseScore"]
first_item["impact"]["baseMetricV3"]["cvssV3"]["baseSeverity"]
```

The new output looks like this:
````
Found 2 known vulnerabilities in 2 packages
Name       Version ID               Fix Versions Severity Score
---------- ------- ---------------- ------------ -------- -----
py         1.11.0  PYSEC-2022-42969              HIGH     7.5
setuptools 65.5.0  PYSEC-2022-43012 65.5.1       MEDIUM   5.9
````

json and cyclonedx output have also been changed so that the severity and score will be visible therein.
Since NVD has a limit of 5 requests over a moving window of 30 seconds, we included the possibility of using an api-key from one's environment, if that is present (as `NVD_API_KEY`). This will increase the limit to 50 requests per 30 seconds, which should cover most use-cases.

A disclaimer for using NVD API has been added, for complying with NVD requests when using their product


matteoannotell:
Closing this PR after [discussion](https://github.com/pypa/pip-audit/issues/654) with maintainers

PR #672, opened 2023-09-08
add note on ignoring unfixed vulnerabilities

### LINKED ISSUE #670: CLI: add a flag to exit with non-zero code only for fixable vulnerabilities

When using pip-audit in CI pipelines, there is no point in ~~failing~~ [exiting with non-zero code] for finding unfixable vulnerability in dependencies.

`--only-fixable` flag (or something similar) to ignore vulnerabilities where no fixed version is available in the database.

A solution would be to run `pip-audit`, trap its exit code. If it fails run `pip-audit --fix` and trap its exit code. If it succeeds, `exit 1`. Otherwise, run `diff` on input and output lists and return its exit code.

**Additional context**

I know that I can trap the exit code all together, filter a single vulnerability ID, or set my CI job to allow failure. But as far as I can see, none of them can replicate this exact behavior.

MFTabriz:
content is based on the discussion with @woodruffw and @di in #670

closes #670

PS: feel free to improve it as you see fit.

### DIFF
--- a/README.md
+++ b/README.md
@@ -468,7 +468,7 @@ exitcode="${?}"
 See [Exit codes](#exit-codes) for a list of potential codes that need handling.
 
 ### Reporting only fixable vulnerabilities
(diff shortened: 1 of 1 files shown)


woodruffw (approved):
LGTM, thank you @MFTabriz!

PR #680, opened 2023-09-26
[DRAFT] Add expiry date with two arguments

KeithWM:
The same functionality as #679, but now with a different argument syntax. I think this is neater, even if the arguments are more verbose. Maybe there's a good synonym for "temporarily ignore" that we could use instead.

### DIFF
--- a/pip_audit/_cli.py
+++ b/pip_audit/_cli.py
@@ -10,8 +10,9 @@ import logging
 import os
 import sys
 from contextlib import ExitStack, contextmanager
+from datetime import datetime
 from pathlib import Path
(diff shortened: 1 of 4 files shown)


pip_audit/_cli.py
```diff
@@ -44,6 +45,8 @@
 package_logger = logging.getLogger("pip_audit")
 package_logger.setLevel(os.environ.get("PIP_AUDIT_LOGLEVEL", "INFO").upper())
 
+DATE_SEP = "::"
+
```
KeithWM, line 49: Left-over from the previous attempt, should be removed.

PR #702, opened 2023-12-05
bump minimum Python to 3.8

woodruffw:
3.7 is EOL, and will soon begin dragging some of our support tooling.


pyproject.toml
```diff
@@ -97,5 +96,4 @@ reset = true
 # Never enforce `E501` (line length violations).
 ignore = ["E501"]
 select = ["E", "F", "I", "W", "UP"]
-target-version = "py37"
```
woodruffw, line 100: NB: no longer needed, since `ruff` checks `requires-python`.

di (approved):

PR #715, opened 2023-12-25
Allow build with cyclonedx-python-lib 6.0.0+

sunpoet:
get_instance() has been renamed to make_outputter() and removed in 6.0.0.

Fixes #705.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -28,7 +28,7 @@ classifiers = [
 ]
 dependencies = [
     "CacheControl[filecache] >= 0.13.0",
-    "cyclonedx-python-lib >= 4,< 6",
+    "cyclonedx-python-lib >= 4,< 7",
     "html5lib>=1.1",
     "packaging>=23.0.0",                 # https://github.com/pypa/pip-audit/issues/464
     "pip-api>=0.0.28",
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -28,7 +28,7 @@ classifiers = [
 ]
 dependencies = [
     "CacheControl[filecache] >= 0.13.0",
-    "cyclonedx-python-lib >= 4,< 6",
+    "cyclonedx-python-lib >= 4,< 7",
```
woodruffw, line 31: If the API in question has been removed, then this bound needs to be changed to:

```suggestion
    "cyclonedx-python-lib >= 6,< 7",
```

woodruffw (changes requested):
LGTM, one suggestion

pyproject.toml
sunpoet: I've changed it to >=5 since make_outputter() was added in https://github.com/CycloneDX/cyclonedx-python-lib/pull/469.

woodruffw: Thanks!

woodruffw (approved):
Thank you @sunpoet!

sunpoet:
Thanks.

Issue #235, opened 2022-02-07
CLI: add a flag for showing each result's alias set


woodruffw:
Follow-on from #232.

One potential format:

```
$ pip-audit --show-dupes
Found 2 known vulnerabilities in 1 package
Name  Version ID                              Fix Versions
----  ------- ------------------------------- ------------
Flask 0.5     PYSEC-2019-179, FOOBAR-2019-666 1.0
Flask 0.5     PYSEC-2018-66                   0.12.3
```

We could also do this as a separate `Aliases` column instead, in which case we might want to make the flag `--show-aliases`.

di:
FWIW the actual flag name I used was just a placeholder, I don't think it's a great name for UX.

woodruffw:
CC @jleightcap: this is a good potential first issue, to get familiar with the codebase.

pboulos:
Hello everyone, I've written an implementation for this as I had need for it in a project I was working on (see #721).

I've looked over your contributing docs, but please let me know if you'd like me to go about this another way (and obviously for any adjustments you'd like to make).

woodruffw:
Thanks @pboulos! I'll give this a review tomorrow. 

CCing @tetsuo-cpp as well 🙂

woodruffw:
Merged with #721! Thanks a ton @pboulos!

PR #721, opened 2024-01-07
Add aliases to output.

pboulos:
Implementation for #235.

### DIFF
--- a/README.md
+++ b/README.md
@@ -171,6 +171,12 @@ optional arguments:
                         defaults to `on` for the `json` format. This flag has
                         no effect on the `cyclonedx-json` or `cyclonedx-xml`
                         formats. (default: auto)
+  --aliases [{on,off,auto}]
+                        includes alias IDs for each vulnerability; `auto`
+                        defaults to `off` for the `column` and `markdown`
+                        formats. This flag has no effect on the
+                        `cyclonedx-json`, `cyclonedx-xml`, and `json`
+                        formats. (default: auto)
   --cache-dir CACHE_DIR
                         the directory to use as an HTTP cache for PyPI; uses
                         the `pip` HTTP cache by default (default: None)
@@ -274,6 +280,16 @@ Flask 0.5     PYSEC-2019-179 1.0
 Flask 0.5     PYSEC-2018-66  0.12.3
 ```
 
+Audit dependencies including aliases:
+```
+$ pip-audit --aliases
+Found 2 known vulnerabilities in 1 package
+Name  Version ID             Fix Versions Aliases
+----  ------- -------------- ------------ ------------------------------------
+Flask 0.5     PYSEC-2019-179 1.0          CVE-2019-1010083,GHSA-5wv5-4vpf-pj6m
+Flask 0.5     PYSEC-2018-66  0.12.3       CVE-2018-1000656,GHSA-562c-5r94-xh97
+```
+
 Audit dependencies including descriptions:
 ```
 $ pip-audit --desc
@@ -295,6 +311,10 @@ Found 2 known vulnerabilities in 1 package
     "vulns": [
       {
         "id": "PYSEC-2019-179",
+        "aliases": [
+          "CVE-2019-1010083",
+          "GHSA-5wv5-4vpf-pj6m"
+        ],
         "fix_versions": [
           "1.0"
         ],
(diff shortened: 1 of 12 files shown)


pip_audit/_format/columns.py
```diff
@@ -131,6 +137,8 @@ def _format_vuln(
         ]
         if applied_fix is not None:
             vuln_data.append(self._format_applied_fix(applied_fix))
+        if self.output_aliases:
+            vuln_data.append(",".join(vuln.aliases))
```
woodruffw, line 141: Nitpick: the columnar format is meant to be human-readable and not machine-readable, so we can add a little bit of whitespace here:

```suggestion
            vuln_data.append(", ".join(vuln.aliases))
```

woodruffw (commented):
LGTM! One tiny nitpick, and needs deconflict.

pip_audit/_cli.py
```diff
@@ -241,6 +267,17 @@ def _parser() -> argparse.ArgumentParser:  # pragma: no cover
...
+        "--aliases",
+        type=VulnerabilityAliasChoice,
+        choices=VulnerabilityAliasChoice,
+        nargs="?",
+        const=VulnerabilityAliasChoice.On,
+        default=VulnerabilityAliasChoice.Auto,
+        help="include aliases for each vulnerability; "
+        "`auto` defaults to `off` for the `column` and `markdown` formats. This flag has no "
+        "effect on the `cyclonedx-json`, `cyclonedx-xml`, or `json` formats.",
+    )
```
woodruffw, line 280: I'm okay with this flag name, assuming others are. CC @tetsuo-cpp @di

README.md
```diff
@@ -171,6 +171,12 @@ optional arguments:
                         defaults to `on` for the `json` format. This flag has
                         no effect on the `cyclonedx-json` or `cyclonedx-xml`
                         formats. (default: auto)
+  --aliases [{on,off,auto}]
+                        includes alias IDs for each vulnerability; `auto`
+                        defaults to `off` for the `column` and `markdown`
+                        formats. This flag has no effect on the
+                        `cyclonedx-json`, `cyclonedx-xml`, and `json`
```
tetsuo-cpp, line 178: What's the reason for forcing this on for the JSON format? It's not a big deal since the alias list won't be large but it's not consistent with the `--desc` flag.

Also if I'm reading this right, the `off` option works identical to `auto` since the JSON format always includes aliases.

I think we should either:
- Allow aliases to be toggled for the JSON format, `auto` defaults to `on` for JSON (this makes the most sense to me).
- Remove the `auto` value.

Thoughts @woodruffw @di?

tetsuo-cpp (commented):
Awesome work @pboulos, just had a question about the behaviour of the `--aliases` flag with the JSON format.

README.md
woodruffw: Yeah, agreed -- I think forcing on (rather than having the default be on) will cause user confusion. I'll make a suggested change below.

pip_audit/_cli.py
```diff
@@ -134,6 +134,32 @@ def __str__(self) -> str:
...
+    Whether or not vulnerability aliases should be added to the `pip-audit` output.
+    """
+
+    On = "on"
+    Off = "off"
+    Auto = "auto"
+
+    def to_bool(self, format_: OutputFormatChoice) -> bool:
+        if format_ is OutputFormatChoice.Json:  # Note: Aliases are always rendered for JSON.
+            return True
```
woodruffw, line 149: Per discussion above: let's remove this, for consistent with the other format-modifying flags.

```suggestion
```

pip_audit/_format/columns.py
pboulos: Yes, this makes sense. To this effect, I've also modified the markdown format file to output in the same manner as it is (as I understand it) also intended to be human readable.

README.md
pboulos: My initial thoughts were that forcing it to be included would be inobtrusive enough that it wouldn't warrant a toggle (as with the case for descriptions).

@tetsuo-cpp the difference between what I previously implemented as `off` and `auto` was that passing `pip-audit ... -f json --aliases off` as a flag would not affect the JSON output, and aliases would have always been included. With `pip-audit ... -f json --desc off`, descriptions would be omitted from the JSON output.

That being said, I agree for the sake of consistency that the behaviour should match `--desc`, and as such have added the necessary code to support that.

tetsuo-cpp (approved):
LGTM, thanks @pboulos.

Issue #732, opened 2024-02-06
PermissionError: [Errno 13] Permission denied: '/tmp/tmpzrluurcb/bin/python3'


rdinoff:
## Bug description

if /tmp/ is mounted with noexec you get the following stack trace.   

## Expected behavior

need an option to put tmp files in a different location 

```
(venv) rdinoff@audit:~/audit$ pip-audit -r requirements.txt
Traceback (most recent call last):
  File "/home/rdinoff/audit/venv/bin/pip-audit", line 8, in <module>
    sys.exit(audit())
  File "/home/rdinoff/audit/venv/lib/python3.9/site-packages/pip_audit/_cli.py", line 509, in audit
    for spec, vulns in auditor.audit(source):
  File "/home/rdinoff/audit/venv/lib/python3.9/site-packages/pip_audit/_audit.py", line 67, in audit
    for dep, vulns in self._service.query_all(specs):
  File "/home/rdinoff/audit/venv/lib/python3.9/site-packages/pip_audit/_service/interface.py", line 154, in query_all
    for spec in specs:
  File "/home/rdinoff/audit/venv/lib/python3.9/site-packages/pip_audit/_dependency_source/requirement.py", line 134, in collect
    yield from self._collect_from_files(collect_files)
  File "/home/rdinoff/audit/venv/lib/python3.9/site-packages/pip_audit/_dependency_source/requirement.py", line 180, in _collect_from_files
    ve.create(ve_dir)
  File "/usr/lib/python3.9/venv/__init__.py", line 75, in create
    self._setup_pip(context)
  File "/usr/lib/python3.9/venv/__init__.py", line 313, in _setup_pip
    subprocess.check_output(cmd, stderr=subprocess.STDOUT)
  File "/usr/lib/python3.9/subprocess.py", line 424, in check_output
    return run(*popenargs, stdout=PIPE, timeout=timeout, check=True,
  File "/usr/lib/python3.9/subprocess.py", line 505, in run
    with Popen(*popenargs, **kwargs) as process:
  File "/usr/lib/python3.9/subprocess.py", line 951, in __init__
    self._execute_child(args, executable, preexec_fn, close_fds,
  File "/usr/lib/python3.9/subprocess.py", line 1837, in _execute_child
    raise child_exception_type(errno_num, err_msg, err_filename)
PermissionError: [Errno 13] Permission denied: '/tmp/tmpzrluurcb/bin/python3'

```

## Platform information

* OS name and version: ubuntu 20.04
[...]

PR #737, opened 2024-02-10
_virtual_env: handle PermissionError

woodruffw:
~~WIP, needs tests.~~

Closes https://github.com/pypa/pip-audit/issues/732

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,12 @@ All versions prior to 0.0.9 are untracked.
 
 ## [Unreleased]
 
+### Fixed
+
+* Improved the error returned to users when their default temporary
+  directory lacks execute permissions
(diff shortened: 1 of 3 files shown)


woodruffw:
Just thinking out loud: this is going to be really annoying to test. I guess we could mock it out, but that won't guarantee that I got the exact callsite right here.

woodruffw:
I looked into this a bit more, and I mis-diagnosed this slightly originally: the error here isn't from our own code, but from `venv` itself: the `_setup_pip` function uses `subprocess` internally, which then explodes for the original `noexec` reason. So current patch is wrong.

Ref: https://github.com/python/cpython/blob/a5f49ce65ef83c41c9cc87566fe59be766d0f3ca/Lib/venv/__init__.py#L357-L360

Issue #662, opened 2023-08-14
Duplicates are not supported in requirements.txt when run with --disable-pip


npalmius:
## Bug description

Duplicate entries (with extras) in a requirements file cause an error when auditing if using the `--disable-pip` option.

This is closely related to #564 (but I would say that it is not a duplicate).

## Reproduction steps

Use the following `requirements.txt` file (generated by poetry):
```
dash==2.7.1 ; python_full_version < "3.9.0" and python_version >= "3.8"
dash[compress]==2.7.1 ; python_full_version < "3.9.0" and python_version >= "3.8"
```

This gives the following outputs:

Without `--disable-pip`:
```
$ pip-audit -r requirements.txt
No known vulnerabilities found
```

With `--disable-pip`:
```
$ pip-audit -r requirements.txt --disable-pip --no-deps
WARNING:pip_audit._cli:The --no-deps flag is redundant when used with --disable-pip
WARNING:pip_audit._cli:--no-deps is supported, but users are encouraged to fully hash their pinned dependencies
WARNING:pip_audit._cli:Consider using a tool like `pip-compile`: https://pip-tools.readthedocs.io/en/latest/#using-hashes
ERROR:pip_audit._cli:package dash has duplicate requirements: dash[compress]==2.7.1 (from RequirementLine(line_number=2, line='dash[compress]==2.7.1 ; python_full_version < "3.9.0" and python_version >= "3.8"', filename=PosixPath('requirements.txt')))
```

For reference, the error is the same if the `requirements.txt` file is generated with hashes, so this is specifically related to `--disable-pip`, not `--no-deps`:
```
$ pip-audit -r requirements.txt --disable-pip
ERROR:pip_audit._cli:package dash has duplicate requirements: dash[compress]==2.7.1 (from RequirementLine(line_number=290, line='dash[compress]==2.7.1 ; python_full_version < "3.9.0" and python_version >= "3.8"     --hash=sha256:0935490e3dd2d2a046291989824732901437bc0336d592eadcb93fd9c866d4c9     --hash=sha256:13cd19d19288a8fb9c2a5cd069c72643758fc2fcda68cacc72474a4ec6b7667d', filename=PosixPath('requirements.txt')))
```

## Expected behavior

`pip-audit -r requirements.txt --disable-pip` should perform the same as `pip-audit -r requirements.txt`

## Platform information

* OS name and version:
* `pip-audit` version: `pip-audit 2.6.1`
* Python version: `Python 3.8.17`
[...]

PR #749, opened 2024-03-17
662 duplicates are not supported in requirements.txt when run with   disable pip

mathbou:
Recently, I run in the same problem described in #662. To avoid this, I propose a finer check for duplicates based on both name and specifier.

As stated in the issue, when the `--disable-pip` flag is used, we could consider that a full requirement resolution has been made. Knowing that, as long as specifiers matches, having duplicates is not a problem. If they don't match, we raise an error like before.

On the side, I also add a small fix for stdout/stderr reading in `pip_audit/_subprocess.py`. I don't know if it's specific to windows, but the fact that a size was specified, I had the process hanging indefinitely.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,11 @@ All versions prior to 0.0.9 are untracked.
 
 ## [Unreleased]
 
+### Fixed
+
+* Improved handling of temporary files on Windows
+  ([#757](https://github.com/pypa/pip-audit/pull/757))
+
 ## [2.7.2]
 
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,15 @@ All versions prior to 0.0.9 are untracked.
 
 ### Fixed
 
-* Auditing a fully-pinned requirements file with `--disable-pip` now allows for duplicates, so long as the duplicates don't have conflicting specifier sets ([#749](https://github.com/pypa/pip-audit/pull/749))
+* Auditing a fully-pinned requirements file with `--disable-pip` now allows for
+  duplicates, so long as the duplicates don't have conflicting specifier sets
+  ([#749](https://github.com/pypa/pip-audit/pull/749))
+
+### Changed
+
+* The default cache locations on macOS and Linux now respect each platform's
+  caching directory idioms (e.g. XDG)
+  ([#814](https://github.com/pypa/pip-audit/pull/814))
 
 ## [2.7.3]
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,10 @@ All versions prior to 0.0.9 are untracked.
 * `pip-audit` now allows some CLI flags to be configured via environment
   variables ([#755](https://github.com/pypa/pip-audit/pull/755))
 
+### Changed
+
+* `--disable-pip` allows duplicate requirements if specifiers matches ([#749](https://github.com/pypa/pip-audit/pull/749))
+
 ## [2.7.3]
 
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,17 +13,17 @@ All versions prior to 0.0.9 are untracked.
 * `pip-audit` now allows some CLI flags to be configured via environment
   variables ([#755](https://github.com/pypa/pip-audit/pull/755))
 
-### Fixed
-
-* Auditing a fully-pinned requirements file with `--disable-pip` now allows for
-  duplicates, so long as the duplicates don't have conflicting specifier sets
-  ([#749](https://github.com/pypa/pip-audit/pull/749))
-
 ### Changed
 
 * The default cache locations on macOS and Linux now respect each platform's
   caching directory idioms (e.g. XDG)
   ([#814](https://github.com/pypa/pip-audit/pull/814))
+
+### Fixed
+
+* Auditing a fully-pinned requirements file with `--disable-pip` now allows for
+  duplicates, so long as the duplicates don't have conflicting specifier sets
+  ([#749](https://github.com/pypa/pip-audit/pull/749))
 
 ## [2.7.3]
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,10 +8,20 @@ All versions prior to 0.0.9 are untracked.
 
 ## [Unreleased]
 
+### Added
+
+* `pip-audit` now allows some CLI flags to be configured via environment
+  variables ([#755](https://github.com/pypa/pip-audit/pull/755))
+
+## [2.7.3]
+
 ### Fixed
 
 * Improved handling of temporary files on Windows
   ([#757](https://github.com/pypa/pip-audit/pull/757))
+
+* Fixed a subprocess deadlock on Windows
+  ([#756](https://github.com/pypa/pip-audit/pull/756))
 
 ## [2.7.2]
 
@@ -589,7 +599,8 @@ All versions prior to 0.0.9 are untracked.
   dependency errors ([#146](https://github.com/pypa/pip-audit/pull/146))
 
 <!-- Release URLs -->
-[Unreleased]: https://github.com/pypa/pip-audit/compare/v2.7.2...HEAD
+[Unreleased]: https://github.com/pypa/pip-audit/compare/v2.7.3...HEAD
+[2.7.3]: https://github.com/pypa/pip-audit/compare/v2.7.2...v2.7.3
 [2.7.2]: https://github.com/pypa/pip-audit/compare/v2.7.1...v2.7.2
 [2.7.1]: https://github.com/pypa/pip-audit/compare/v2.7.0...v2.7.1
 [2.7.0]: https://github.com/pypa/pip-audit/compare/v2.6.3...v2.7.0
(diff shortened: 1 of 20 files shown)


woodruffw:
Thanks for the patch @mathbou! I'll review this today.

pip_audit/_dependency_source/requirement.py
```diff
@@ -225,18 +225,23 @@ def _fix_file(self, filename: Path, fix_version: ResolvedFixVersion) -> None:
...
                 and (req.marker is None or req.marker.evaluate())
                 and req.req is not None
             ):
-                if req.name in req_names:
+                duplicate_req_specifier = req_specifiers.get(req.name)
+
+                if not duplicate_req_specifier:
+                    req_specifiers[req.name] = req.specifier
+
+                elif duplicate_req_specifier != req.specifier:
```
woodruffw, line 241: Just making sure I understand the behavior here: rather than deduping on just the requirement's `name` (i.e. package name), we're now deduping on the conflicting specifiers for duplicated names, right?

IIUC that means that we'll still have some false positives here (since `SpecifierSet != SpecifierSet` does not imply that the two are incompatible), right? If so that's fine, just making sure I understand 🙂

pip_audit/_subprocess.py
```diff
@@ -52,9 +52,8 @@ def run(args: Sequence[str], *, log_stdout: bool = False, state: AuditState = Au
     # once `stdout` hits EOF, so we don't have to worry about that blocking.
     while not terminated:
         terminated = process.poll() is not None
-        # NOTE(ww): Buffer size chosen arbitrarily here and below.
-        stdout += process.stdout.read(4096)  # type: ignore
-        stderr += process.stderr.read(4096)  # type: ignore
+        stdout += process.stdout.read()  # type: ignore
+        stderr += process.stderr.read()  # type: ignore
```
woodruffw, line 56: This is for a separate bug you're seeing, right? If so, I'd prefer to break this out into a separate issue and PR -- we've had trouble with this bit of code before, so I don't want to make a change here as part of a separate bugfix 🙂

pip_audit/_dependency_source/requirement.py
mathbou: > rather than deduping on just the requirement's name (i.e. package name), we're now deduping on different specifiers for duplicated names

Yep, that's the idea. As long as each package sharing the same name, also share the same specifier, it is considered identical.

> IIUC that means that we'll still have some false positives here (since SpecifierSet != SpecifierSet does not imply that the two are incompatible), right? If so that's fine, just making sure I understand 🙂

I'm not sure how we could still get false positives here 🤔 Have you a specific case in mind ? 
Unless you modify the resolved requirements, it is very unlikely to get different `SpecifierSet` for the package and its extras.

pip_audit/_subprocess.py
mathbou: No problem, it's here now #756

pip_audit/_dependency_source/pyproject.py
```diff
@@ -142,8 +142,8 @@ def fix(self, fix_version: ResolvedFixVersion) -> None:
             # Now dump the new edited TOML to the temporary file.
             toml.dump(pyproject_data, tmp)
 
-            # And replace the original `pyproject.toml` file.
-            os.replace(tmp.name, self.filename)
+        # And replace the original `pyproject.toml` file.
+        os.replace(tmp.name, self.filename)
```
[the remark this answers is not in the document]
mathbou: Looks like an oversight, it should also be split. 

I moved it out because I had several tests failing with: `PermissionError: [WinError 32] The process cannot access the file because it is being used by another process`. When the `os.replace` is called, the file is still open for `r+w` and Windows don't like it.

pip_audit/_dependency_source/requirement.py
woodruffw: > I'm not sure how we could still get false positives here 🤔 Have you a specific case in mind ?
Unless you modify the resolved requirements, it is very unlikely to get different SpecifierSet for the package and its extras.

What I'm thinking of is something like this:

```
example >= 1.0
example >= 1.0, < 3.0
```

In this case, `SpecifierSet(">= 1.0") != SpecifierSet(">= 1.0, < 3.0")`, but the two are not incompatible (all versions in 1.x and 2.x satisfy both sets). Unless I'm misunderstanding, I believe this change will still incorrectly reject this as a "duplicate" dependency, when one is a superset of the other.

pip_audit/_subprocess.py
woodruffw: Thank you!

pip_audit/_dependency_source/requirement.py
mathbou: You are correct, it this case it will be rejected.
Maybe it would be more appropriate to reject them as `incompatible` instead of `duplicate`, what do you think ?

mathbou:
It's been a while here, is there anything that prevent us to go further with this PR ? @woodruffw

woodruffw:
> It's been a while here, is there anything that prevent us to go further with this PR ? @woodruffw

Nope, I've just been delayed in reviews, sorry 😅. I'll do another pass on this today.

(Thank you very much for keeping this PR alive and conflict-free!)

pip_audit/_dependency_source/requirement.py
woodruffw: > Maybe it would be more appropriate to reject them as incompatible instead of duplicate, what do you think ?

Actually, I think I misunderstood the context here, sorry! From a quick look I think we can never reach this kind of double-range state, since `--disable-pip` means that we always have a fully-pinned requirements file.

So I think we can continue to describe this as `duplicate`.

woodruffw (commented):
Thanks @mathbou, this looks good to me!

Could you add a CHANGELOG entry describing the bugfix here? The other entries in the file should serve as a reference for our preferred entry format 🙂

mathbou:
@woodruffw I updated the changelog, feel free to change it if it's not clear enough

CHANGELOG.md
```diff
@@ -13,6 +13,10 @@ All versions prior to 0.0.9 are untracked.
 * `pip-audit` now allows some CLI flags to be configured via environment
   variables ([#755](https://github.com/pypa/pip-audit/pull/755))
 
+### Changed
```
woodruffw, line 16: This is a bugfix:

```suggestion
### Fixed
```

CHANGELOG.md
```diff
@@ -13,6 +13,10 @@ All versions prior to 0.0.9 are untracked.
...
+
+* `--disable-pip` allows duplicate requirements if specifiers matches ([#749](https://github.com/pypa/pip-audit/pull/749))
```
woodruffw, line 18:
```suggestion
* Auditing a fully-pinned requirements file with `--disable-pip` now allows for duplicates, so long as the duplicates don't have conflicting specifier sets ([#749](https://github.com/pypa/pip-audit/pull/749))
```

woodruffw (changes requested):

pip_audit/_dependency_source/requirement.py
```diff
@@ -313,12 +318,21 @@ def _collect_preresolved_deps(
...
+            # but different specifiers, meaning a badly resolved requirements.txt
+            elif duplicate_req_specifier != req.specifier:
                 raise RequirementSourceError(
                     f"package {req.name} has duplicate requirements: {str(req)}"
                 )
-            req_names.add(req.name)
+            else:
+                # We have a duplicate requirement for the same package and the specifier matches
+                # As they would return the same result from the audit, there no need to yield it a second time.
+                continue
```
woodruffw, line 335: From CI: this line needs coverage!

mathbou: This is weird, coverage complains about this line only with py3.8 and 3.9, from 3.10 and above I get 100%.

woodruffw: This is probably a manifestation of https://github.com/nedbat/coveragepy/issues/198, which was ultimately resolved in 3.10+ via PEP 626: https://peps.python.org/pep-0626/

The two solutions here are either to drop the `else: ... continue` *or* add a coverage ignore comment.

mathbou: Unfortunately, dropping the `else` cant be done without also changing the surrounding code, it's part of the logic.

I add an ignore comment, hope it will suffice 🤞

pip_audit/_dependency_source/requirement.py
```diff
@@ -313,12 +318,21 @@ def _collect_preresolved_deps(
...
+            # but different specifiers, meaning a badly resolved requirements.txt
+            elif duplicate_req_specifier != req.specifier:
                 raise RequirementSourceError(
                     f"package {req.name} has duplicate requirements: {str(req)}"
                 )
-            req_names.add(req.name)
+            else:
+                # We have a duplicate requirement for the same package and the specifier matches
+                # As they would return the same result from the audit, there no need to yield it a second time.
+                continue  # pragma: no cover
```
woodruffw, line 335: Sorry, but I think we need to do better than skipping coverage on this -- it's a core part of the change, so we can't just skip it. 

This PR has been open for a long time (that's my fault), so I'll take over here and try to rewrite it into a shape that `coverage.py` understands.

mathbou: Does it really worth that much effort ? Especially as coverage does not complain on py3.10+, and technically the problem come from coverage itself 🤔

woodruffw: Hmm, good point. I'll update the comment to say we can drop the ignore after we're 3.10+ and after that this LGTM.

woodruffw (approved):

woodruffw:
Thanks for your patience here @mathbou!

PR #755, opened 2024-03-30
Allow environment variables to set CLI defaults

### LINKED ISSUE #754: Provide environment variables for CLI option defaults
### Pre-submission checks

- [X] I am **not** reporting a new vulnerability or requesting a new vulnerability identifier. These **must** be reported or managed via upstream dependency sources or services, not this repository.
- [X] I agree to follow the [PSF Code of Conduct](https://www.python.org/psf/conduct/).
- [X] I have looked through the open issues for a duplicate request.

### What's the problem this feature will solve?

Currently, we can only use command line options to configure behavior of `pip-audit`. There is no way to configure its behavior once and retain the configuration for follow-up runs.

### Describe the solution you'd like

Every CLI option should have its default value configurable via an environment variable (following an easy-to-remember naming scheme). Using a CLI option should take precedence over any such value.

This way, in a terminal one could use `export PIP_AUDIT_...=...`. In the YAML file of popular CI systems one could set environment variables in the respective `ENVIRONMENT` section or similar.

### Example Implementation

```python
parser.add_argument("-f", "--format", default=os.environ.get("PIP_AUDIT_FORMAT", "columns"))
parser.add_argument("-o", "--output", default=os.environ.get("PIP_AUDIT_OUTPUT", "stdout"))
parser.add_argument("--progress-spinner", default=os.environ.get("PIP_AUDIT_PROGRESS_SPINNER", "on"))
parser.add_argument("--timeout", default=os.environ.get("PIP_AUDIT_TIMEOUT", "15"))
```

### Additional context

This suggestion would be complementary to #694, but quicker, easier and more straight-forward to implement as it is without the complexity of reading a configuration file.

bittner:
Adds a few environment variable getters that allow to override the default value of some CLI options.

Fixes #754

### DIFF
--- a/test/test_cli.py
+++ b/test/test_cli.py
@@ -1,5 +1,8 @@
+from pathlib import Path
+
 import pretend  # type: ignore
 import pytest
+from cli_test_helpers import ArgvContext, EnvironContext
 
 import pip_audit._cli
 from pip_audit._cli import (
@@ -215,3 +218,21 @@ def test_print_format(monkeypatch, vuln_count, pkg_count, skip_count, print_form
         pass
 
     assert bool(dummyformat.format.calls) == print_format
+
+
+def test_environment_variable():
+    """Environment variables set before execution change CLI option default."""
+    with ArgvContext("pip-audit"), EnvironContext(
+        PIP_AUDIT_DESC="off",
+        PIP_AUDIT_FORMAT="markdown",
+        PIP_AUDIT_OUTPUT="stderr",
+        PIP_AUDIT_PROGRESS_SPINNER="off",
+        PIP_AUDIT_VULNERABILITY_SERVICE="osv",
+    ):
+        args = pip_audit._cli._parser().parse_args()
+
+    assert args.desc == VulnerabilityDescriptionChoice.Off
+    assert args.format == OutputFormatChoice.Markdown
+    assert args.output == Path("stderr")
+    assert not args.progress_spinner
+    assert args.vulnerability_service == VulnerabilityServiceChoice.Osv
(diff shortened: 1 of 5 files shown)


bittner:
@woodruffw I've not yet covered documenting the environment variables. Not sure what the best way would be to do it. A multi-line explanation in [`argparse.ArgumentParser(description)`](https://github.com/pypa/pip-audit/blob/main/pip_audit/_cli.py#L197) could be one possibility.

test/test_cli.py
```diff
@@ -215,3 +218,21 @@ def test_print_format(monkeypatch, vuln_count, pkg_count, skip_count, print_form
...
+def test_environment_variable():
+    """Environment variables set before execution change CLI option default."""
+    with ArgvContext("pip-audit"), EnvironContext(
+        PIP_AUDIT_DESC="off",
+        PIP_AUDIT_FORMAT="markdown",
+        PIP_AUDIT_OUTPUT="stderr",
+        PIP_AUDIT_PROGRESS_SPINNER="off",
+        PIP_AUDIT_VULNERABILITY_SERVICE="osv",
+    ):
+        args = pip_audit._cli._parser().parse_args()
```
woodruffw, line 232: It'll be a little less clean, but my preference here would be to use `pytest`'s built-in monkeypatching (via the `monkeypatch` fixture) for argv/environment patching here, rather than adding an additional test-only dependency.

woodruffw:
Thanks @bittner, this looks like a great initial set!

> I've not yet covered documenting the environment variables. Not sure what the best way would be to do it.

Hmm, IMO it'd be fine to document it in the README alone, i.e. under the "Usage" section in a new sub-section above the "Exit codes" one: <https://github.com/pypa/pip-audit?tab=readme-ov-file#exit-codes>

(A short explainer that some flags honor the environment + a bullet list of each of them would be perfect.)

woodruffw (approved):

PR #756, opened 2024-04-01
Fix pipe reading hanging indefinitely on Windows

mathbou:
While doing tests with pip-audit, I noticed several times the process stalling while installing the isolated env. 

By removing the stdout/stderr read size, it seems to run properly, however it's quite tricky to reproduce.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,11 @@ All versions prior to 0.0.9 are untracked.
 
 ## [Unreleased]
 
+### Fixed
+
+* Improved handling of temporary files on Windows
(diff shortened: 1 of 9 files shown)


woodruffw:
I'm acknowledging this PR so you know I'm not ignoring it, but JFYI: I'm pretty backlogged at the moment, so I can't guarantee a timely review here 🙂

woodruffw:
Took a look, and I have no major objection to removing the explicit buffer sizes. That being said, it'd be ideal to have a better understanding of _why_ this would deadlock on Windows: we set `bufsize=0` so the pipe should be unbuffered regardless of OS, meaning that neither `read()` should ever block.

woodruffw (approved):

PR #757, opened 2024-04-01
Many tests failing on Windows

mathbou:
While testings for others PRs, I get a lot of `permissionError`. Windows does not like moving/replacing files if there aren't properly closed or another process is still on it.

I also get normalization problems on paths used in tests, so I enforced a posix format instead of a basic string cast.

### DIFF
--- a/pip_audit/_dependency_source/requirement.py
+++ b/pip_audit/_dependency_source/requirement.py
@@ -193,7 +193,7 @@ class RequirementSource(DependencySource):
             # Make temporary copies of the existing requirements files. If anything goes wrong, we
             # want to copy them back into place and undo any partial application of the fix.
             tmp_files: list[IO[str]] = [
-                stack.enter_context(NamedTemporaryFile(mode="w")) for _ in self._filenames
+                stack.enter_context(NamedTemporaryFile(mode="r+")) for _ in self._filenames
             ]
             for filename, tmp_file in zip(self._filenames, tmp_files):
                 with filename.open("r") as f:
@@ -278,10 +278,9 @@ class RequirementSource(DependencySource):
     def _recover_files(self, tmp_files: list[IO[str]]) -> None:
         for filename, tmp_file in zip(self._filenames, tmp_files):
             try:
-                os.replace(tmp_file.name, filename)
-                # We need to tinker with the internals to prevent the file wrapper from attempting
-                # to remove the temporary file like in the regular case.
-                tmp_file._closer.delete = False  # type: ignore[attr-defined]
+                tmp_file.seek(0)
+                with filename.open("w") as f:
+                    shutil.copyfileobj(tmp_file, f)
             except Exception as e:
                 # Not much we can do at this point since we're already handling an exception. Just
                 # log the error and try to recover the rest of the files.
--- a/pip_audit/_dependency_source/requirement.py
+++ b/pip_audit/_dependency_source/requirement.py
@@ -5,7 +5,6 @@ Collect dependencies from one or more `requirements.txt`-formatted files.
 from __future__ import annotations
 
 import logging
-import os
 import re
 import shutil
 from contextlib import ExitStack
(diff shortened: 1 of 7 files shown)


pip_audit/_dependency_source/requirement.py
```diff
@@ -220,7 +219,7 @@ def _fix_file(self, filename: Path, fix_version: ResolvedFixVersion) -> None:
         #
         # This time we're using the `RequirementsFile.parse` API instead of `Requirements.from_file`
         # since we want to access each line sequentially in order to rewrite the file.
-        reqs = list(RequirementsFile.parse(filename=str(filename)))
+        reqs = list(RequirementsFile.parse(filename=filename.as_posix()))
```
woodruffw, line 222: Confirming: is `as_posix` okay for Windows here?

test/dependency_source/test_requirement.py
```diff
@@ -378,7 +379,9 @@ def test_requirement_source_fix_rollback_failure(monkeypatch, req_file):
     def mock_replace(*_args, **_kwargs):
         raise OSError
 
-    monkeypatch.setattr(os, "replace", mock_replace)
+    from tempfile import _TemporaryFileWrapper
+
+    monkeypatch.setattr(_TemporaryFileWrapper, "seek", mock_replace, raising=False)
```
woodruffw, line 384: This seems slightly off: we're replacing `seek` with a mocked `os.replace` implementation, it looks like.

mathbou: I agree, it's not ideal, but as it is related to [this part](https://github.com/pypa/pip-audit/pull/757/files#diff-a05f9f5ebfa6898a1b210166fc630ca5558f90eb0b6358267f66288f27bbdc8aL281-R282), which is also new, it's the only way I found to keep the test behavior intact

pip_audit/_dependency_source/requirement.py
mathbou: It's completely transparent. The `as_posix` only ensure that the path uses `/` instead of `\`.

test/dependency_source/test_requirement.py
```diff
@@ -121,13 +122,13 @@ def test_requirement_source_git(req_file):
         [
             (
                 req_file(),
-                "git+https://github.com/unbit/uwsgi.git@1bb9ad77c6d2d310c2d6d1d9ad62de61f725b824",
+                "git+https://github.com/benoitc/gunicorn.git@61ccfd6c38d477a908e0f376757bbb884438053a",
```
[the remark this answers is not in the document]
woodruffw: Hmm, that config setting makes sense, but `core/logging.` doesn't _look_ like an invalid NTFS path. 

Given that this is probably going to continue being a source of weirdness in the future, let's use an official example:

```suggestion
                "git+https://github.com/pypa/sampleproject.git@5d277956b5a571dac16b28db74e5f2b780d9af5f",
```

(Corresponding to https://github.com/pypa/sampleproject/commit/5d277956b5a571dac16b28db74e5f2b780d9af5f)

pip_audit/_dependency_source/requirement.py
woodruffw: Okay. I couldn't remember if POSIX paths were fine on Windows or not, but if so this looks fine 🙂

test/dependency_source/test_requirement.py
woodruffw: I understand that! I'm saying that it should be renamed to `mock_seek` since that's what it's mocking 😉

woodruffw:
A few small nits; otherwise this is looking good! Thanks for your work here @mathbou!

test/dependency_source/test_requirement.py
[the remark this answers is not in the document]
mathbou: > but core/logging. doesn't look like an invalid NTFS path

I don't understand either, I suspect the faulty character doesn't even appear in the log 🤷‍♂️ 

> let's use an official example:

I didn't know there was a sample project, it's way better indeed 👍

test/dependency_source/test_requirement.py
mathbou: oh yes, sorry, I didn't realize you were talking about the variable name 😅 I changed it.

woodruffw (approved):
LGTM overall!

@mathbou would you mind adding a `CHANGELOG` entry as well? Something under "fixed" probably that matches the style of other entries 🙂

Issue #794, opened 2024-07-09
Use pip's `http-v2` cache


woodruffw:
With https://github.com/pypa/pip/pull/11143, `pip` uses `$CACHE/http-v2` instead of `$CACHE/http`. This new directory uses a separated body format, rather than collecting the cached responses into a single serialized file each (both header and body).

As a result, `pip-audit` is currently only pulling from the older `http` cache, which will slowly go away in user installs over time. We should switch `_cache.py` to mimic and reuse the new `http-v2` cache.

More context: here is the current `SafeFileCache` implementation within `pip`: https://github.com/pypa/pip/blob/34e70ba5964824ac8c52773c76c658923ee36194/src/pip/_internal/network/cache.py#L32

xovishnukosuri:
I have a fix for this. pip 23.3 introduced the `http-v2` cache directory, but `_get_pip_cache()` in `_cache.py` still always appends `/http` to the pip cache base path. On any system running pip 23.3+, the `http` directory doesn't exist and pip-audit silently gets no cache benefit.

The fix is a version check: select `http-v2` for pip ≥ 23.3, `http` for older pip.

**Changes to `pip_audit/_cache.py`:**

```python
# pip 23.3 switched from a single-file `http` cache to a two-file `http-v2` cache that stores
# response body separately from metadata to reduce peak memory usage.
# See: https://github.com/pypa/pip/issues/11149
_PIP_HTTP_V2_VERSION = Version("23.3")
```

```python
def _get_pip_cache() -> Path:
    ...
    cache_dir = process.stdout.decode("utf-8").strip("\n")
    # pip 23.3+ uses `http-v2` instead of `http` as the cache subdirectory name.
    http_subdir = "http-v2" if _PIP_VERSION >= _PIP_HTTP_V2_VERSION else "http"
    return Path(cache_dir) / http_subdir
```

**Changes to `test/test_cache.py`:**
- Update `test_get_pip_cache` to assert `cache_dir.stem in {"http", "http-v2"}` (the old assertion `== "http"` would fail on pip 23.3+)
- Add `test_get_pip_cache_http_v2` that monkeypatches `_PIP_VERSION = Version("23.3")` and asserts `http-v2`
- Add `test_get_pip_cache_http_legacy` that monkeypatches `_PIP_VERSION = Version("23.2")` and asserts `http`

[...]

PR #814, opened 2024-08-17
More respectful convention for caching

### LINKED ISSUE #342: Caching: Use a more respectful default location?
At the moment, we fall back to `~/.pip-audit/cache` if `pip cache dir` fails. This isn't a "respectful" default, since it doesn't adhere to either XDG (which probably wants us to consult `XDG_CACHE_HOME`) or macOS's conventions (`~/Library/Caches`, probably).

This probably isn't the most important change to make since 99% of users won't notice it anyways, but it would make us a slightly better participant in the user's home directory.

aanghelidi:
A PR that aims to close #342.

All the logic is inside `_get_internal_cache_path` in `_cache.py`.

### DIFF
--- a/pip_audit/_cache.py
+++ b/pip_audit/_cache.py
@@ -6,6 +6,7 @@ from __future__ import annotations
 
 import logging
 import os
+import platform
 import subprocess
 import sys
 from pathlib import Path
@@ -28,7 +29,24 @@ _MINIMUM_PIP_VERSION = Version("20.1")
 
 _PIP_VERSION = Version(str(pip_api.PIP_VERSION))
 
-_PIP_AUDIT_INTERNAL_CACHE = Path.home() / ".pip-audit-cache"
+
+def _get_internal_cache_path() -> Path:
+    """
+    Returns the default cache directory path used internally by `pip-audit`.
+
+    This should be used as a fallback when other caching options are not specified.
+
+    On macOS, it will be in `$HOME/Library/Caches/pip_audit`.
+
+    On other OS, it will follow the `XDG Base Directory Specification` and
+    place the cache in either `$HOME/.cache/pip-audit` or respect `$XDG_CACHE_HOME`
+    environment variable i-e `$XDG_CACHE_HOME/pip-audit`.
+    """
+    return (
+        (Path.home() / "Library" / "Caches" / "pip-audit")
+        if platform.system() == "Darwin"
+        else (Path(os.getenv("XDG_CACHE_HOME", Path.home() / ".cache")) / "pip-audit")
+    )
 
 
 def _get_pip_cache() -> Path:
@@ -60,6 +78,9 @@ def _get_cache_dir(custom_cache_dir: Path | None, *, use_pip: bool = True) -> Pa
     if custom_cache_dir is not None:
         return custom_cache_dir
 
+    # Retrieve pip-audit's default internal cache path following OS conventions.
+    pip_audit_cache_dir = _get_internal_cache_path()
+
     # Respect pip's PIP_NO_CACHE_DIR environment setting.
     if use_pip and not os.getenv("PIP_NO_CACHE_DIR"):
         pip_cache_dir = _get_pip_cache() if _PIP_VERSION >= _MINIMUM_PIP_VERSION else None
@@ -68,11 +89,11 @@ def _get_cache_dir(custom_cache_dir: Path | None, *, use_pip: bool = True) -> Pa
         else:
             logger.warning(
                 f"pip {_PIP_VERSION} doesn't support the `cache dir` subcommand, "
-                f"using {_PIP_AUDIT_INTERNAL_CACHE} instead"
+                f"using {pip_audit_cache_dir} instead"
             )
-            return _PIP_AUDIT_INTERNAL_CACHE
+            return pip_audit_cache_dir
     else:
-        return _PIP_AUDIT_INTERNAL_CACHE
+        return pip_audit_cache_dir
 
 
 class _SafeFileCache(FileCache):
--- a/pip_audit/_cache.py
+++ b/pip_audit/_cache.py
@@ -48,14 +48,6 @@ def _get_pip_cache() -> Path:
     return http_cache_dir
 
 
-def _delete_legacy_cache_dir(current_cache_dir: Path, legacy_cache_dir: Path) -> None:
-    """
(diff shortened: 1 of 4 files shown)


woodruffw (approved):
Thanks @aanghelidi, this looks pretty reasonable to me. `pip-audit`'s choice of cache location isn't a stability commitment we make, so we should be fine to change this without making a major release.

A couple of thoughts:

* Maybe we could use [`platformdirs`](https://pypi.org/project/platformdirs/) to abstract the cache location discovery for us? OTOH the current logic in `_get_internal_cache_path` is pretty simple; curious what you think.
* Maybe we should blow away the old cache if a new path is discovered? Otherwise, users will end up silently keeping `_PIP_AUDIT_INTERNAL_CACHE` around in their `$HOME` taking up space, which isn't very polite of us.

For the second point, something like this is what I have in mind:

```python
# if the selected internal cache isn't the old one, delete the old one.
if internal_cache_path != _PIP_AUDIT_LEGACY_INTERNAL_CACHE:
    shutil.rmtree(_PIP_AUDIT_LEGACY_INTERNAL_CACHE)
```

CC @di for thoughts as well 🙂

aanghelidi:
Thanks @woodruffw  for the review 🙂 TIL about [platformdirs](https://pypi.org/project/platformdirs/) ! This is indeed a really handy package. Regarding the aforementioned points:

- There is several pros about using `platformdirs`:
  - simplicity. Using `platformdirs`, the implementation of `pip-audit`'s cache location is a one-liner.
  - we do not reinvent the wheel. This means less maintenance for the project.
  - consistency. Upon research, it seems that `platformdirs` is already used in several `pypa` projects:
    -  [hatch](https://github.com/pypa/hatch/tree/master): see [here](https://github.com/pypa/hatch/blob/e07d2984aba528262616b70891860e5aefb12775/src/hatch/config/model.py#L414)
    - [pipx](https://github.com/pypa/pipx/tree/main): see [here](https://github.com/pypa/pipx/blob/4462299fa391b7349074facc39c73767853c9d02/src/pipx/paths.py#L118)
    - [setuptools](https://github.com/pypa/setuptools): see [here](https://github.com/pypa/setuptools/tree/main/setuptools/_vendor/platformdirs) (as a vendored library)

- Regarding the potential con(s):
  - dependency cost. We add one more dependency to the project. Although `platformdirs` itself has no dependencies, it's still a decision to make. Maybe this is not worth adding given the simplicity of `_get_internal_cache_path` implementation (which I have the feeling could be simplified).
  
- Regarding `PIP_AUDIT_INTERNAL_CACHE`, it is indeed not nice to keep the old cache in `$HOME`. Removing it seems to be a good idea. Maybe, as an alternative, using `shutil.move` to move the old cache to the new location could be an idea?

I added a commit using `platformdirs`  for the cache location, let me know what you think 🙂

woodruffw:
> * Regarding `PIP_AUDIT_INTERNAL_CACHE`, it is indeed not nice to keep the old cache in `$HOME`. Removing it seems to be a good idea. Maybe, as an alternative, using `shutil.move` to move the old cache to the new location could be an idea?

I think we should just blow it away for now -- #794 covers some larger architectural changes we need to make to handle newer versions of `pip`'s cache anyways, so these old paths may not be super useful when moved to the new location.

> I added a commit using platformdirs for the cache location, let me know what you think 🙂

LGTM, thank you!

aanghelidi:
> I think we should just blow it away for now -- https://github.com/pypa/pip-audit/issues/794 covers some larger architectural changes we need to make to handle newer versions of pip's cache anyways, so these old paths may not be super useful when moved to the new location

Thank you for the context @woodruffw ! I added a extra commit to delete the old cache that goes into that direction. Let me know what you think 🙂

pip_audit/_cache.py
```diff
@@ -46,6 +48,14 @@ def _get_pip_cache() -> Path:
     return http_cache_dir
 
 
+def _delete_legacy_cache_dir(current_cache_dir: Path, legacy_cache_dir: Path) -> None:
+    """
+    Deletes the legacy `pip-audit` if it exists.
+    """
+    if current_cache_dir != legacy_cache_dir:
+        shutil.rmtree(legacy_cache_dir)
```
woodruffw, line 56: No need for a helper here; let's just inline this into `_get_cache_dir` IMO.

aanghelidi: The last commit now inline the deletion into `_get_cache_dir` :heavy_check_mark:

woodruffw (approved):
Thanks @aanghelidi, LGTM!

woodruffw (commented):
Sorry, one last request: could you add a `CHANGELOG` entry for this change? See the other entries in the log for our preferred style.

woodruffw (approved):

hseg:
Tests fails on my machine -- with `$XDG_CACHE_HOME=~/.local/var/cache/`, I get
```
WARNING  pip_audit._cache:_cache.py:81 pip 1.0.0 doesn't support the `cache dir` subcommand, using /home/gesh/.local/var/cache/pip-audit instead
FAILED test/test_cache.py::test_get_cache_dir_do_not_use_pip[on Linux] - AssertionError: assert PosixPath('/home/gesh/.local/var/cache/pip-audit') == PosixPath('/home/gesh/.c...
FAILED test/test_cache.py::test_get_cache_dir_pip_disabled_in_environment[on Linux] - AssertionError: assert PosixPath('/home/gesh/.local/var/cache/pip-audit') == PosixPath('/home/gesh/.c...
FAILED test/test_cache.py::test_get_cache_dir_old_pip[on Linux] - AssertionError: assert PosixPath('/home/gesh/.local/var/cache/pip-audit') == PosixPath('/home/gesh/.c...
```
which is odd since I have
```
$ pip --version
pip 25.0 from /usr/lib/python3.13/site-packages/pip (python 3.13)
```

hseg:
Judging by the log output and the testcases, I suspect what's going on is you've
correctly implemented the "check XDG variables and fallback" logic, but are not
accounting for it in the test cases.
In that case, it'd be enough to unset `$XDG_CACHE_HOME` in the three
above-mentioned tests (and add a test exercising the expected behaviour that if
`$XDG_CACHE_HOME` is set but `pip` cannot be relied on to give the path the
obtained cache directory is indeed `$XDG_CACHE_HOME/pip-audit` -- perhaps by
setting `$XDG_CACHE_HOME` to a canary value)

PR #851, opened 2024-11-04
workflows: address zizmor findings, add zizmor workflow

woodruffw:
Addresses some low findings from `zizmor` + adds a security scanning workflow that'll catch anything else that pops up, proactively.


di (approved):

di:
@woodruffw Many merge conflicts here.

PR #869, opened 2024-12-21
Add junit output format

bibi21000:
Add junit output format. Junit is a well known format used in CIs

### DIFF
--- a/pip_audit/_format/junit.py
+++ b/pip_audit/_format/junit.py
@@ -1,0 +1,210 @@
+"""
+Functionality for formatting vulnerability results as an array of JSON objects.
+"""
+
+from __future__ import annotations
+
+from datetime import datetime
+from typing import cast
+
+import pip_audit._fix as fix
+import pip_audit._service as service
+
+from .interface import VulnerabilityFormat
+
+
+class JunitFormat(VulnerabilityFormat):
+    """
+    An implementation of `VulnerabilityFormat` that formats vulnerability results as a
+    junit xml file (https://github.com/testmoapp/junitxml).
+    Junit is a wellknow and usefull format when working with Continuous Integration.
+    """
+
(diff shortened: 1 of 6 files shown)


pip_audit/_format/junit.py
```diff
@@ -0,0 +1,210 @@
...
+            nb_skipped += sk
+            nb_failures += fa
+            nb_errored += er
+            output += outp
+        output_xml = '<?xml version="1.0" encoding="UTF-8"?>\n'
+        output_xml += f'<testsuites name="pip-audit" tests="{nb_tests}" failures="{nb_failures}" '
+        output_xml += f'skipped="{nb_skipped}" errors="{nb_errored}" timestamp="{timestamp}">\n'
+        output_xml += output
+        output_xml += '</testsuites>'
+        return output_xml
```
woodruffw, line 73: I'm sorry, but this approach is a hard blocker: we can't be manually building up XML with string concatenation, *especially* without any proper escaping.

woodruffw (changes requested):
I'm not fundamentally opposed to adding JUnit XML as a supported output format, but it would have been really nice to have a discussion thread before this PR was opened. In particular, one outcome of that discussion would probably have been identifying an appropriate output construction format, such as a new dependency or a generic XML document builder.

In lieu of that discussion thread:

* Is there an XML document building package or, even better, a JUnit XML package we can use?
* Can you provide some details on why you want/need this output format? 
* Is there a reason why you can't use `--format=json` and adapt the output to JUnit XML via an external tool?

bibi21000:
I needed it ... i did it ...
Why Junit ? one of the most format used in CI ... and perfect for this kind of report.
As far as i know ... there is no official xsd for junit ... using xml tools could be a mess.
This Junit is used on Jenkins and works fine.
Ofc ... it's possible to convert json to xml ... all is possible

PR #880, opened 2025-02-10
test-cache: Fix XDG tests, make tests more consistent

hseg:
Solving the issues I raised in
https://github.com/pypa/pip-audit/pull/814#issuecomment-2646555094, this commit
set (at least on my machine) fixes the tests on Linux systems with custom XDG
variables.

While I was there, I noticed the tests were a little inconsistent, so this PR
also contains a couple of commits addressing that. Three of the changes I'm
unsure about, so I'd appreciate a second look at them before this is merged.

- **fix(test_cache): Correct XDG testing logic**

  The tests on Linux were relying on XDG_CACHE_HOME being set to its fallback
  value of `∼/.cache`, which made them fail on systems with custom values for
  `XDG_CACHE_HOME`.

  To account for this, we take the cue from the Windows tests and set
  `XDG_CACHE_HOME` to a known, custom, value. This incidentally also tests that
  our logic can account for custom values of `XDG_CACHE_HOME`, so this doesn't
  need its own test.

  The value of `/tmp/home/.cache` was chosen as a cross between the Windows
  value of `/tmp/AppData/Local` and the fallback `∼/.cache`

  (This commit is a MVP for the PR, the others can be ignored if desired)

  Fixes: #814

- **RFC: fix(test_cache): Make tests more consistent**
  - `test_get_cache_dir`: Follow all other tests in not casting to posix paths
    in checking paths are as expected, instead checking an equal Path object is
    roundtripped. NOTE: This edit I'm least confident in, not least since I
    don't have a Windows machine available to test this on.
  - `test_get_cache_dir_old_pip`: Remove logic checking whether _get_cache_dir
    accepts explicitly-set paths. It is duplicated from `test_get_cache_dir`,
    doesn't appear to be exercising any new codepath, and it is unclear why this
    case needs extra exercise
  - `test_get_pip_cache`, `test_get_cache_dir_old_pip`: follow the lead of
    `test_get_cache_dir_pip_disabled_in_environment` and inline the call to
    `_get_cache_dir`
- **RFC: fix(test_cache): Cross-OS test_get_cache_dir**

  In view of my doubts above in re whether I broke `test_get_cache_dir` on
  Windows, run the test simulating all platforms. However, I am unsure that my
  understanding of `pathlib` and the relevant monkeypatching is correct, or if
  this commit is even needed.

### DIFF
--- a/test/test_cache.py
+++ b/test/test_cache.py
@@ -20,8 +20,11 @@ def _patch_platformdirs(monkeypatch: MonkeyPatch, sys_platform: str) -> None:
     # as cache definition is stored in the top level `__init__.py` file of the
     # `platformdirs` package
     importlib.reload(platformdirs)
+    # Setting directory-controlling environment variables to known state
     if sys_platform == "win32":
         monkeypatch.setenv("LOCALAPPDATA", "/tmp/AppData/Local")
+    elif sys_platform == "linux":
+        monkeypatch.setenv("XDG_CACHE_HOME", "/tmp/home/.cache")
 
 
 def test_get_cache_dir(monkeypatch):
@@ -48,7 +51,7 @@ def test_get_pip_cache():
     [
         pytest.param(
             "linux",
-            Path.home() / ".cache" / "pip-audit",
+            Path("/tmp") / "home" / ".cache" / "pip-audit",
             id="on Linux",
         ),
         pytest.param(
@@ -76,7 +79,7 @@ def test_get_cache_dir_do_not_use_pip(monkeypatch, sys_platform, expected):
     [
         pytest.param(
             "linux",
-            Path.home() / ".cache" / "pip-audit",
+            Path("/tmp") / "home" / ".cache" / "pip-audit",
             id="on Linux",
         ),
         pytest.param(
@@ -105,7 +108,7 @@ def test_get_cache_dir_pip_disabled_in_environment(monkeypatch, sys_platform, ex
     [
         pytest.param(
             "linux",
-            Path.home() / ".cache" / "pip-audit",
+            Path("/tmp") / "home" / ".cache" / "pip-audit",
             id="on Linux",
         ),
         pytest.param(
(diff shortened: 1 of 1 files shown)


woodruffw (approved):
Thanks @hseg, these changes look great! Please undraft this when you think it's ready and I'll merge it.

hseg:
OK, removed RFC markers from commit messages. I'd appreciate you running the new
`test_get_cache_dir` on a Windows machine, as I'm not confident my
monkeypatching accurately tests that behaviour on my Linux machine.
But other than that, I'm happy with this one.

woodruffw:
Sorry for the delay here! I've added a Windows CI job, and it looks like it indeed needs some tweaks for actual Windows hosts. If you're able to make those tweaks then please do; otherwise I'll push some fixes here when I have the free time.

hseg:
Good that you added a windows runner - these seem to have been broken even without my changes - but I have no idea at first glance why it's failing. I don't have the time to investigate right now, but I'll see if I can sometime next week.

27 feb 2025 17:58:31 William Woodruff ***@***.***>:

woodruffw:
OK, it looks like the only real remaining issue here is with `_patch_platformdirs` in the cache tests -- doing this kind of `sys.platform` monkeypatching is probably too brittle, since CPython itself uses `sys.platform` to conditionally use/import all kinds of other things.

Rather than monkeypatching here, we'll probably want to rewrite these tests to condition on each runtime OS, so that macOS tests for macOS, Windows for Windows, etc. I can look into that in a bit.

woodruffw:
Okay, this was exceedingly annoying to fix: I needed to refactor how we do coverage collection since Windows tests don't have 100% coverage (on purpose). We now assert 100% test coverage over all test runs combined, which is what we really wanted anyways.

To get there I needed to refactor our testing setup a decent bit: I needed to remove `pytest-cov` since it does coverage combination implicity, and needed to enable pytest's "parallel" mode _without_ adding `pytest-xdist`, since the latter messes with coverage collection without some additional hacks. There are some follow-ons from there, but that should all be good enough to merge here.

Thanks again @hseg, and sorry for the wait!

PR #888, opened 2025-04-02
PEP 751 support

woodruffw:
~~WIP.~~

Some scattered thoughts:

- Right now this adds a `--locked` flag to enable lockfile collection. Does this make sense? Should it happen by default instead, taking priority over `pyproject.toml` when present?
- Right now this collects all `pylock.*.toml` files by default, not just the generic one or a particular service-specific one. Does this make sense? Should `--locked` take a value to control this, e.g. `--locked=all` for the current behavior and `--locked=<service>` for just `pylock.<service>.toml`?
- ~~This only checks `[[packages]]`. I need to do another closer read of PEP 751 to understand if there are other parts of the file we should collect from.~~ Checked, and `packages` should be the only part.
- ~~This doesn't perform any deduplication at the moment, i.e. `foo==1.2.3` will be audited multiple times if specified in multiple lockfiles or multiple times in the same file (which the PEP allows). This probably won't happen often but I should probably add that deduplication, similarly to how `requirements.txt` inputs are handled.~~ This is moot, since deduplication is done at the audit layer.
- This currently skips any package that doesn't have a version. I _think_ this is probably the right behavior, but perhaps it should be stricter, i.e. skip if it's a non-sdist/wheel but fail/warn if an sdist or wheel is missing a version?


pip_audit/_dependency_source/pylock.py
```diff
@@ -0,0 +1,112 @@
...
+        """
+        Raises `NotImplementedError` if called.
+
+        We don't support fixing dependencies in lockfiles, since
+        lockfiles should be managed/updated by their packaging tool.
+        """
+
+        raise NotImplementedError(
+            "lockfiles cannot be fixed directly; use your packaging tool to perform upgrades"
+        )
```
di, line 100: This makes me wish that including the command/configuration that generated the lockfile was standardized as well...

di (approved):

pip_audit/_dependency_source/pylock.py
woodruffw: Yeah, that would have been nice! `created-by` gets pretty close and might be helpful to include in the error message, so I'll look at working that in.

PR #903, opened 2025-05-19
Implement the Ecosyste.ms vulnerability backend

tnytown:
Resolves #894.

PR #904, opened 2025-05-20
Implement helper to create `VulnerabilityResult`

tnytown:
Depends on #903.

This takes `ids`, `description`, `fix_versions`, and `published` and
splits `ids` into `id` and aliases as appropriate internally. Priority
is given to PYSEC and CVE identifiers for `id`.


woodruffw (approved):
Thanks @tnytown!

PR #951, opened 2025-10-01
feat: add support for environment variable `PIP_AUDIT_IGNORE_VULN` to override the command-line option of the same name

### LINKED ISSUE #948: Feature: new environment variable PIP_AUDIT_IGNORE_VULN
### Pre-submission checks

- [x] I am **not** reporting a new vulnerability or requesting a new vulnerability identifier. These **must** be reported or managed via upstream dependency sources or services, not this repository.
- [x] I agree to follow the [PSF Code of Conduct](https://www.python.org/psf/conduct/).
- [x] I have looked through the open issues for a duplicate request.

### What's the problem this feature will solve?

CI breaks when new vulnerabilities are reported, and it’s tedious to modify CI to add & ignore vulnerabilities. Having an environment variable would make managing ignoring vulnerabilities easier and more flexible.

### Describe the solution you'd like

A new environment variable `PIP_AUDIT_IGNORE_VULN` which mimics the behavior of the `--ignore-vuln` command line option. The environment variable could take a space-separated list of vulnerability ids.

### Additional context

_No response_

jenstroeger:
Closes #948

Leaving this as a draft for now: I’ll need to add tests and documentation, but this one-liner might suffice? The value for the new environment variable would be a space separated list of vulnerabilities.

I’m a bit torn about the name: the command-line argument can be repeated so `--ignore-vuln` (singular) makes sense. In contrast, the environment variable takes one or more IDs so it should be `PIP_AUDIT_IGNORE_VULNS` — but that would conflict with the command-line arg’s name. Thoughts?


woodruffw:
> In contrast, the environment variable takes one or more IDs so it should be PIP_AUDIT_IGNORE_VULNS — but that would conflict with the command-line arg’s name. Thoughts?

I'm okay with plural -- we don't need perfect correspondence with the flag when doing so would be confusing to users (like here with plurals) 🙂

woodruffw:
One thing that's worth checking here is to see how the environment behaves when the user also passes options as flags: I would expect the flags to take precedence (and for the environment to be ignored), but we should confirm that argparse does that instead of something weird in `append` mode.

jenstroeger:
I updated the PR.

> One thing that's worth checking here is to see how the environment behaves when the user also passes options as flags: I would expect the flags to take precedence (and for the environment to be ignored), but we should confirm that argparse does that instead of something weird in `append` mode.

Alas, that’s not how `append` behaves: the command line args are appended to the default list provided by the environment variable ([docs](https://docs.python.org/3/library/argparse.html#action)). If that’s ok then I think it ought to be documented properly, or else if we want to change that behavior then we’d add a custom [Action](https://docs.python.org/3/library/argparse.html#argparse.Action) to change `append`’s default behavior.

PR #1038, opened 2026-05-18
feat: trust CAs when contacting ESMS, OSV and PyPi

### LINKED ISSUE #345: --trusted-host  or  --cert as in pip itself

Private repositories might have alternative certificates. 
Easily using an extra or alternative index, such as with --extra-index-url of pip-audit might thus fail, if pip-audit can't find that.

Pip itself has the following optional arguments it can use. Pip-audit having the same functionality would be nice. 

  --trusted-host <hostname>   Mark this host or host:port pair as trusted, even though it does not have valid or any
                              HTTPS.
  --cert <path>               Path to PEM-encoded CA certificate bundle. If provided, overrides the default. See 'SSL
                              Certificate Verification' in pip documentation for more information.

guerda:
This pull request enables trusting locally installed Certification Authorities so that pip-audit works in corporate networks, where custom CAs are used to issue SSL certificates for all domains. Those certificates are not trusted by the usual CAs and therefore `pip-audit` fails to check for CVEs.

Implementation
------------------------
* I added [`truststore`](https://pypi.org/projects/truststorie) as a dependency to access all CAs which are installed on the local operating system
* Inject all CAs installed on the system into SSL context, so that request to ESMS, OSV and PyPi can be validated against custom CAs.
* add changelog entry for custom CA trust

Tests and style
---------------------
I tested the feature behind a firewall where custom CAs are used for all websites.
All unit tests which run on main work also on this branch. There's one failing which is also failing on main.
I linted and reformatted the whole project

Documentation
-----------------------
I added inline documentation and proper commit messages. I also added a line for the feature in the CHANGELOG

Fixes #345 
Please let me know if you have any feedback on this.

### DIFF
--- a/pip_audit/_service/esms.py
+++ b/pip_audit/_service/esms.py
@@ -11,7 +11,6 @@ from typing import Any, cast
 from urllib.parse import urlencode
 
 import requests
-import truststore
 from packaging.specifiers import SpecifierSet
 from packaging.version import Version
 
@@ -49,9 +48,6 @@ class EcosystemsService(VulnerabilityService):
         `timeout` is an optional argument to control how many seconds the component should wait for
         responses to network requests.
         """
-        # Injecting all certificate authorities installed on the system into SSL
-        # context for all HTTP calls.
-        truststore.inject_into_ssl()
         self.session = caching_session(cache_dir, use_pip=False)
         self.timeout = timeout
 
--- a/pip_audit/_service/esms.py
+++ b/pip_audit/_service/esms.py
@@ -11,6 +11,7 @@ from typing import Any, cast
 from urllib.parse import urlencode
 
 import requests
+import truststore
 from packaging.specifiers import SpecifierSet
 from packaging.version import Version
 
@@ -48,6 +49,9 @@ class EcosystemsService(VulnerabilityService):
         `timeout` is an optional argument to control how many seconds the component should wait for
         responses to network requests.
         """
+        # Injecting all certificate authorities installed on the system into SSL
+        # context for all HTTP calls.
+        truststore.inject_into_ssl()
         self.session = caching_session(cache_dir, use_pip=False)
         self.timeout = timeout
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -22,11 +22,11 @@ jobs:
           - "3.14"
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+      - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
         with:
           persist-credentials: false
 
-      - uses: actions/setup-python@a309ff8b426b58ec0e2a45f0f869d46889d02405 # v6.2.0
+      - uses: actions/setup-python@ece7cb06caefa5fff74198d8649806c4678c61a1 # v6.3.0
         with:
           python-version: ${{ matrix.python }}
           cache: "pip"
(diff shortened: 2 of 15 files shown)


guerda:
Hi @woodruffw !
I'd like to hear your thoughts about this feature to support custom CAs via truststore in pip-audit.
Thanks in advance for considering.

pip_audit/_service/esms.py
```diff
@@ -48,6 +49,9 @@ def __init__(
         `timeout` is an optional argument to control how many seconds the component should wait for
         responses to network requests.
         """
+        # Injecting all certificate authorities installed on the system into SSL
+        # context for all HTTP calls.
+        truststore.inject_into_ssl()
```
woodruffw: I'm not sure it makes sense to inject this at each service's initialization, rather than once at process start. The examples I've seen from `truststore` suggest that it should be invoked exactly once, ideally very early in program initialization (to ensure that no possible network routes are left uncovered by it).

woodruffw (commented):
Thanks @guerda! I'm not opposed to adding this, but I think it should be called exactly once somewhere early in initialization rather than once per service.

Separately: maybe this should be behind a flag? I'm not sure; pip had it behind a flag for a while, but IIRC it's been the default for the last few releases. Curious if you have opinions on that.

pip_audit/_service/esms.py
guerda: I agree, I searched for one central position to inject the CAs into the context.
As far as I understood pip-audit's architecture, all services (pypi, osv and esms) construct their own sessions.

I can try it out by injecting it more centrally, where would that be? I assume that In the CLI it won't catch everything?

guerda:
> Thanks @guerda! I'm not opposed to adding this, but I think it should be called exactly once somewhere early in initialization rather than once per service.

I agree, I would like to see it once in the whole package, but haven't found the position yet.

> Separately: maybe this should be behind a flag? I'm not sure; pip had it behind a flag for a while, but IIRC it's been the default for the last few releases. Curious if you have opinions on that.

I considered a flag or a separate packaging (with extras like pip-audit[truststore]). As pip is using it as default and truststore is pretty lightweight, I would suggest to include it as default.

woodruffw:
> I agree, I would like to see it once in the whole package, but haven't found the position yet.

I think right at the top of `audit()` would be fine, since that's our primary entrypoint:

https://github.com/pypa/pip-audit/blob/d6720db2c6aeef231459c5d227fcc92eb2f7a9f5/pip_audit/_cli.py#L441



> I would suggest to include it as default.

OK, good enough for me 🙂

guerda:
> I think right at the top of `audit()` would be fine, since that's our primary entrypoint:
> 
> https://github.com/pypa/pip-audit/blob/d6720db2c6aeef231459c5d227fcc92eb2f7a9f5/pip_audit/_cli.py#L441

I will try that out and update the PR if it works.

guerda:
Injecting it generally in the `audit()` function does not work. 12 test cases fail. I think it's because each validation service creates an own request session.

I found another entry point: the caching session in `_cache.py`. 
https://github.com/pypa/pip-audit/blob/main/pip_audit/_cache.py#L172

I should be able to inject the SSL context from truststore into this caching session which is used by all three validators.

woodruffw:
> Injecting it generally in the `audit()` function does not work. 12 test cases fail. I think it's because each validation service creates an own request session.

Hmm, that doesn't make a lot of sense -- the whole point of `inject_into_ssl` is that it's a global operation, so _any_ request session (or other HTTP object that uses `ssl` under the hood) should receive the mutated global state.

(Unless you're saying that it's because the unit tests that fail don't pass through `audit()` first? That's plausible if so, in which case maybe it should be even higher up, like in the top-level `__init__.py`. I'm not sure how idiomatic that is though.)

guerda:
> (Unless you're saying that it's because the unit tests that fail don't pass through `audit()` first? That's plausible if so, in which case maybe it should be even higher up, like in the top-level `__init__.py`. I'm not sure how idiomatic that is though.)

I assume that this is the case. I added it into `__init__.py` and with that, tests are green again and it works.

guerda:
What do you think, would adding the injection into `__init__.py` be acceptable or should I do that in `audit()` and then in the test fixture, too?

guerda:
@woodruffw I now pushed a version where I injected the CAs only in the `__init__.py`. It works for the tests and the CLI.

Let me know what you think.

woodruffw:
Sorry, I haven't had time to re-review this. I'll try and get to it in the coming week.

guerda:
Hi @woodruffw !
Just want to check in if you have the chance to review this added feature. I resolved the conflicts with the main branch.

Issue #573, opened 2023-03-23
_subprocess.py stdout reading may corrupt UTF-8 characters, and then fail when decodes the data


dnso86:
## Bug description

`_subprocess.py` may not read the whole contents of the `stdout` and `stderr` pipes. [Attempting to decode](https://github.com/pypa/pip-audit/blob/main/pip_audit/_subprocess.py#L68) the resulting incomplete `stdout` data as UTF-8 might lead to an **UnicodeDecodeError**.

### mock_pip.py

Generates output with a two-byte UTF-8 character split "in half" at the buffer boundary.

```
import random
import string
import sys

from nose.tools import assert_raises

BUFFER_LENGTH = 4096


def random_text(length):
    characters = string.ascii_lowercase + string.digits
    return ''.join(random.choice(characters) for i in range(length)).encode()

if __name__ == '__main__':
    # COPYRIGHT SIGN (U+00A9)
    two_byte_utf8 = b'\xc2\xae'

    assert two_byte_utf8.decode('utf-8')

    buffer_length_text = random_text(BUFFER_LENGTH)
    buffer_length_plus_one_byte = random_text(BUFFER_LENGTH - 1) + two_byte_utf8

    # It is valid UTF-8
    assert buffer_length_plus_one_byte.decode('utf-8')

    # One iteration of while wouldn't read it all
    assert len(buffer_length_plus_one_byte) > BUFFER_LENGTH

    buffer_length_broken_utf_8 = buffer_length_plus_one_byte[:BUFFER_LENGTH]
    one_byte_broken_utf_8 = buffer_length_plus_one_byte[BUFFER_LENGTH:]

    with assert_raises(UnicodeDecodeError):
        buffer_length_broken_utf_8.decode('utf-8')

    with assert_raises(UnicodeDecodeError):
        one_byte_broken_utf_8.decode('utf-8')

    sys.stdout.buffer.write(buffer_length_text)
    sys.stdout.flush()

    sys.stdout.buffer.write(buffer_length_broken_utf_8)
    sys.stdout.flush()

    # This will not be read
    sys.stdout.buffer.write(one_byte_broken_utf_8)
    sys.stdout.flush()
```

### subprocess_run_isolated.py

Minimal reproducible example based on `_subprocess.py`:

```
import subprocess

if __name__ == '__main__':
    process = subprocess.Popen(
        [
            'python3',
            'mock_pip.py',
        ],
        bufsize=0,
        stdout=subprocess.PIPE,
[...]

PR #1051, opened 2026-06-03
_subprocess: decode only complete UTF-8 codepoints in progress output

### LINKED ISSUE #574: Subprocess: fix stream handling
#572 provides a temporary fix by doing codepoint substitution, but #573 demonstrates that our underlying poll/buffering technique is unsound: our buffering might split a unicode sequence, resulting in invalid encodings in an otherwise valid stream. 

Rather than performing the poll-buffer loop ourselves, we should probably use `Popen.communicate()`.

arpitjain099:
Fixes #574.

#572 fixed the crash on the final decode, but as #573 showed, the
poll-and-read loop in `_subprocess.run` is still unsound for the live
progress output. The loop reads `stdout` unbuffered, so a read can land in
the middle of a multi-byte UTF-8 sequence. Each iteration then decoded the
whole accumulated buffer with `errors="replace"`, which turns a codepoint
that is actually valid (once the rest of its bytes arrive on the next read)
into a U+FFFD replacement character in the progress stream.

Following the maintainer's note on the issue, the right fix is to stop
decoding partial byte sequences. Rather than switch wholesale to
`Popen.communicate()` (which would drop the incremental progress updates
that this wrapper exists to provide), this keeps the poll/read loop and its
responsiveness intact and feeds each chunk through a
`codecs.getincrementaldecoder("utf-8")`. The incremental decoder buffers an
incomplete trailing sequence until it is complete, so the progress output
only ever sees fully-decoded codepoints. `final=True` is passed once the
stream hits EOF to flush any genuinely-truncated bytes, preserving the
previous replace behaviour for malformed input. The final result decode is
unchanged.

### Testing

Added a regression in `test/test_subprocess.py` that drives `run` with a
scripted process whose stdout splits the Euro sign (U+20AC, `e2 82 ac`)
across two reads, so the first read carries an incomplete codepoint. It
asserts that no progress update contains a replacement character and that
the final output is exactly `€`. The test fails on the previous code (the
first progress update is `�`) and passes with this change. A second
end-to-end test runs a real subprocess emitting the codepoint and confirms
`run` returns it intact.

`make lint` (ruff format/check, mypy, interrogate, typos) is clean.

### Changelog

Added an entry under `[Unreleased]` / `Fixed`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,13 @@ All versions prior to 0.0.9 are untracked.
 
 ## [Unreleased]
 
+### Fixed
+
+* Fixed live progress output emitting replacement characters when a subprocess's
+  stdout split a multi-byte UTF-8 sequence across reads. The poll-buffer loop now
+  decodes incrementally and only surfaces complete codepoints
+  ([#574](https://github.com/pypa/pip-audit/issues/574))
+
 ## [2.10.0]
 
 ### Added
--- a/pip_audit/_subprocess.py
+++ b/pip_audit/_subprocess.py
@@ -3,6 +3,7 @@ A thin `subprocess` wrapper for making long-running subprocesses more
 responsive from the `pip-audit` CLI.
 """
 
+import codecs
 import os.path
 import subprocess
 from collections.abc import Sequence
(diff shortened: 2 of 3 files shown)


di:
@noezhiya-dot Please point your LLM somewhere else, thanks.

PR #1057, opened 2026-06-11
docs: scope pre-commit hook to requirements file

### LINKED ISSUE #334: Better pre-commit hook

The current suggested pre-commit hook [here](https://github.com/trailofbits/pip-audit#pre-commit-support) will run on every commit.

```yaml
  - repo: https://github.com/trailofbits/pip-audit
    rev: v2.4.3
    hooks:
      -   id: pip-audit
          args: ["-r", "requirements.txt"]
```

Check the suggested hook for `pip-tools`: https://github.com/jazzband/pip-tools#version-control-integration

```yaml
  - repo: https://github.com/jazzband/pip-tools
    rev: 6.3.0
    hooks:
      - id: pip-compile
        files: ^requirements/production\.(in|txt)$
        args: [--index-url=https://example.com, requirements/production.in]
```

This runs only when the relevant files have changed.

What do you think of updating the suggested pre-commit config to use a `files` filter?

That way, `pip-audit` is less invasive, and we cut down on CI compute costs. 🌎 

**Additional context**

Himanshuagrawal4:
### Summary
- add a `files` filter to the documented pre-commit hook example
- explain that the pattern should match the dependency file being audited

### Why
This keeps the hook from running on every commit when the audited requirements file was not changed.

Fixes #334.

### Checks
- `git diff --check`

### DIFF
--- a/README.md
+++ b/README.md
@@ -112,12 +112,16 @@ For example, using `pip-audit` via `pre-commit` to audit a requirements file:
     hooks:
       -   id: pip-audit
           args: ["-r", "requirements.txt"]
+          files: ^requirements\.txt$
 
 ci:
   # Leave pip-audit to only run locally and not in CI
   # pre-commit.ci does not allow network calls
   skip: [pip-audit]
 ```
+
+The `files` pattern keeps the hook focused on changes to the requirements file being
+audited. Adjust the pattern if your project stores dependency files elsewhere.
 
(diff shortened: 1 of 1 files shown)


woodruffw:
This doesn't seem right, given that there are multiple other ways to invoke pip-audit that don't involve requirements.txt.

PR #1074, opened 2026-07-02
Fix subprocess stream decoding for split UTF-8 output

### LINKED ISSUE #574: Subprocess: fix stream handling

0xTaoZ:
Fixes #574.

This updates the subprocess wrapper to drain stdout and stderr concurrently while decoding progress stdout with an incremental UTF-8 decoder. That prevents split multibyte sequences from being surfaced as replacement characters while preserving the existing progress update behavior.

Tests added:
- split multibyte stdout with `log_stdout=True`
- split multibyte stderr on subprocess failure

Local checks:
- `.venv/bin/python -m pytest test/test_subprocess.py -q`
- `.venv/bin/python -m ruff check pip_audit/_subprocess.py test/test_subprocess.py`
- `.venv/bin/python -m mypy pip_audit/_subprocess.py`
- `.venv/bin/python -m ruff format --check pip_audit/_subprocess.py test/test_subprocess.py`
- `git diff --check`

### DIFF
--- a/pip_audit/_subprocess.py
+++ b/pip_audit/_subprocess.py
@@ -5,7 +5,11 @@ responsive from the `pip-audit` CLI.
 
 import os.path
 import subprocess
+import threading
+import time
+from codecs import getincrementaldecoder
 from collections.abc import Sequence
+from io import BufferedReader
 from subprocess import Popen
 
 from ._state import AuditState
@@ -24,6 +28,14 @@ class CalledProcessError(Exception):
         self.stderr = stderr
 
 
+def _read_stream(stream: BufferedReader, output: bytearray) -> None:
+    """
+    Read a subprocess stream into the given output buffer.
+    """
+    while chunk := stream.read(8192):
+        output.extend(chunk)
+
+
 def run(args: Sequence[str], *, log_stdout: bool = False, state: AuditState = AuditState()) -> str:
     """
     Execute the given arguments.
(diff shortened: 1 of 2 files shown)


noqt (commented):
NOQT tested current head `e689b73e`. The two new UTF-8 tests pass, but there’s a race in the live progress path that can permanently skip output.

`bytes(stdout[stdout_decoded_len:])` takes a snapshot, then `stdout_decoded_len = len(stdout)` reads the still-growing buffer separately. If the reader thread appends between those operations, the offset advances past bytes that were never fed into the incremental decoder. The final return value is intact, but `state.update_state()` loses those bytes.

I reproduced that interleaving deterministically: a child emitted `A`, `B`, then `C`; `run()` returned `ABC`, while the final progress log was only `C`.

Could you advance by the captured chunk length instead (`chunk = bytes(...)`; decode it; `stdout_decoded_len += len(chunk)`), or lock the snapshot and offset update, and add this interleaving as a regression? The two added tests, Ruff, format, mypy, and `git diff --check` otherwise pass in NOQT’s independent run.

